PR #7, opened 2018-05-23
fix readme dependency

weissi:
Motivation:

in SwiftPM it's either of these three formats

- from: "0.1.0"
- .exact("0.1.0")
- .upToNextMajor(from: "0.1.0")

and I messed up the last one (forgot the `from:`)

Modifications:

fixed Package.swift

Result:

recommended syntax actually works


normanmaurer:
late LGTM

PR #9, opened 2018-07-14
FixedLengthFrameDecoder

FranzBusch:
Implement a decoder for frames with a fixed length.  https://github.com/apple/swift-nio/pull/479

Motivation:
Sometimes messages come in fixed length frames. https://github.com/apple/swift-nio/issues/474

Modifications:
Added a FixedLengthFrameDecoder.

Result:
A decoder for messages with a fixed length.


Sources/NIOExtras/FixedLengthDecoder.swift
```diff
@@ -0,0 +1,67 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+/// Error indicating that after an operation some unused bytes are left.
+public struct LeftOverBytesError: Error {
```
FranzBusch: @Lukasa I didn't know where to put this Error. Should I create a separate file?

Lukasa: Yeah, a separate file. In fact, we have a weird pattern developing for this, which you can see in our [HTTP/2 repository](https://github.com/apple/swift-nio-http2/blob/master/Sources/NIOHTTP2/HTTP2Error.swift). Essentially, you create a top-level enum with no cases, and make this struct a member of that. You then also create a tag protocol and conform this struct to it.

Lukasa: That said, for this you may do better to just make this error a nested type in `FixedLengthFrameDecoder`.

Lukasa:

PR #11, opened 2018-08-03
Adds a line-based frame decoder that can split received buffers on line endings.

ldewailly:
Adding ability to split received buffer on line endings.

### Motivation:

As per [#473](https://github.com/apple/swift-nio/issues/473).

### Modifications:

Added a new decoder (`LineBasedFrameDecoder`) that splits incoming buffers on line end characters.

### Result:

Received buffers will be split on line end character(s) (`'\n'` or `'\r\n'`), with these characters
stripped in the resulting buffers.


Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,86 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+///
```
Lukasa: Nit: remove this leading empty comment line.

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,86 @@
...
+    public typealias InboundIn = ByteBuffer
+    public typealias InboundOut = ByteBuffer
+    public var cumulationBuffer: ByteBuffer?
+    // we keep track of the last scan end index if we didn't find the delimiter
+    private var lastIndex: ByteBufferView.Index? = nil
+    
+    public init() { }
+    
+    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
+        if let frame = try findNextFrame(buffer: &buffer) {
```
Lukasa: Nit: please use the explicit `self`.

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+    
+    private func findNextFrame(buffer: inout ByteBuffer) throws -> ByteBuffer? {
+        var view = buffer.readableBytesView
+        // get the view's true start, end indexes
+        let _startIndex = view.startIndex
+        let _endIndex = view.endIndex
+        // start where we left off last scan or from the beginning
+        let firstIndex = min(self.lastIndex ?? _startIndex, _endIndex)
```
Lukasa: It is not safe to assume the indices of a `ByteBuffer` are static across invocations of `decode`. Instead of saving the explicit index, you should save the offset of that index from the `readerIndex`. That offset will stay valid.

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,86 @@
...
+        view = view.dropFirst(firstIndex)
```
Lukasa: This is not going to work either: this works only if the `readerIndex` is `0`.

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,86 @@
...
+                let dropCarriageReturn = length > 0
+                        && buffer.getBytes(at: buffer.readerIndex + length - 1, length: 1) == [0x0D]
+                let buff = buffer.readSlice(length: dropCarriageReturn ? length - 1 : length)
+                // drop the delimiter (and trailing carriage return if appicable)
+                buffer.moveReaderIndex(forwardBy: dropCarriageReturn ? 2 : 1)
+                // reset the last scan start index since we found a line
+                self.lastIndex = nil
+                return buff
+            }
+            view = view.dropFirst(1)
```
Lukasa: I don't love this loop. How about this?

```swift
while let carriageReturnIndex = view.firstIndex(of: 0x0a) {
    // Do work
    view = view.suffix(from: view.index(after: carriageReturnIndex))
}
```

(Some of these methods may be 4.2 beta and either need back porting or replacing by something similar, but I think it much better communicates the purpose of the loop.)

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,86 @@
...
+        let _endIndex = view.endIndex
+        // start where we left off last scan or from the beginning
+        let firstIndex = min(self.lastIndex ?? _startIndex, _endIndex)
+        view = view.dropFirst(firstIndex)
+        while !view.isEmpty {
+            if view.starts(with: "\n".utf8) {
+                let length = view.startIndex - _startIndex
+                // check if the line ends with carriage return, if so drop it
+                let dropCarriageReturn = length > 0
+                        && buffer.getBytes(at: buffer.readerIndex + length - 1, length: 1) == [0x0D]
```
Lukasa: This is weird. Maybe keep a view of all of the readable bytes separate so we can just index it to ask this question?

Sources/NIOExtras/LineBasedFrameDecoder.swift
Lukasa: In fact, it's probably a good idea to write a test that does this. It's easily done: write a test that has one and a half lines in the byte buffer, such that it returns the first and then partially completes the second. Then, call `discardReadBytes` on the cumulation buffer and save it back to the handler, before you write the rest of the data in.

Sources/NIOExtras/LineBasedFrameDecoder.swift
weissi: this is fairly inefficient. We allocate a new array here just to check one character. `getInteger(at: buffer.readerIndex + length - 1, as: UInt8.self) == 0xd` is more efficient

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,86 @@
...
+                return buff
+            }
+            view = view.dropFirst(1)
+        }
+        // next scan we start where we stopped
+        self.lastIndex = max(0, _endIndex - 1)
+        return nil
+    }
+    
+    public func handlerRemoved(ctx: ChannelHandlerContext) {
```
weissi: could we handle this on _both_ `channelInactive` and `handlerRemoved`? We could have a new `private func handleLeftOvers(...) { ... }` which does this work which is then just called from both `handlerRemoved` and `channelInactive`. The reason is that when `handlerRemoved` is being invoked, the pipeline is likely currently being deconstructed already and other handlers might have already (partly) cleaned up.

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,86 @@
...
+import NIO
+
+///
+/// A decoder that splits incoming `ByteBuffer`s around line end
+/// character(s) (`'\n'` or `'\r\n'`).
+///
+/// Let's, for example, consider the following received buffer:
+///
+///     +----+-------+------------+
+///     | AB | C\nDE | F\r\nGHI\n |
```
weissi: should we make it configurable what sequences work as splitting sequence?

Sources/NIOExtras/LineBasedFrameDecoder.swift
ldewailly: @Lukasa @weissi is this approach efficient enough?

```swift
if let delimitorIndex = view.firstIndex(of: 0x0A) { // '\n'
    let length = delimitorIndex - _startIndex
    let dropCarriageReturn = delimitorIndex > 0 && view[delimitorIndex - 1] == 0x0D // '\r'
    let buff = buffer.readSlice(length: dropCarriageReturn ? length - 1 : length)
    // drop the delimiter (and trailing carriage return if appicable)
    buffer.moveReaderIndex(forwardBy: dropCarriageReturn ? 2 : 1)
    return buff
}
```

ldewailly: So we do `view.firstIndex(of: 0x0A)` instead of the loop, and `view[delimitorIndex - 1]` to check for trailing CR character

Lukasa: Yes, that should be just fine. Note that, here's a nit, you want "delimiter", not "delimitor".

Sources/NIOExtras/LineBasedFrameDecoder.swift
ldewailly: @weissi I started down this path but thought we can do a few optimisations knowing the delimiter is a single character (`\n`). Happy to generalise this decoder if you feel this would be more useful. Though I suspect `\n` or `\r\n` will represent the majority of use cases

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,97 @@
...
+    }
+    
+    private func findNextFrame(buffer: inout ByteBuffer) throws -> ByteBuffer? {
+        var view = buffer.readableBytesView
+        // get the view's true start, end indexes
+        let _startIndex = view.startIndex
+        // start where we left off or from the beginning
+        let lastIndex = self.lastScanOffset.flatMap { buffer.readerIndex + $0 }
+        let firstIndex = min(lastIndex ?? _startIndex, view.endIndex)
+        view = view.dropFirst(firstIndex - buffer.readerIndex)
```
Lukasa: Looks like we're doing some unnecessary math above. We're persisting the offset from the reader index, extracting the new reader index, doing two optionality checks, and a `min` function, and then subtracting the reader index again, all to answer the question "how far past the reader index should we go?"

However, we already knew the answer: it's `self.lastScanOffset ?? 0`. So lines 53 through 59 can be replaced by `var view = buffer.readableBytesView.dropFirst(self.lastScanOffset ?? 0)`.

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,97 @@
...
+        // look for the delimiter
+        if let delimiterIndex = view.firstIndex(of: 0x0A) { // '\n'
+            let length = delimiterIndex - _startIndex
```
Lukasa: If we make the change above, `_startIndex` can be replaced by `buffer.readerIndex`.

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,97 @@
...
+            let dropCarriageReturn = delimiterIndex > 0 && view[delimiterIndex - 1] == 0x0D // '\r'
+            let buff = buffer.readSlice(length: dropCarriageReturn ? length - 1 : length)
+            // drop the delimiter (and trailing carriage return if appicable)
+            buffer.moveReaderIndex(forwardBy: dropCarriageReturn ? 2 : 1)
+            // reset the last scan start index since we found a line
+            self.lastScanOffset = nil
+            return buff
+        }
+        // next scan we start where we stopped
+        self.lastScanOffset = buffer.readerIndex + view.count
```
Lukasa: The `lastScanOffset` is equal to `buffer.readableBytes`.

Sources/NIOExtras/LineBasedFrameDecoder.swift
Lukasa: Oh in fact, we don't even need the optional folding, because we can now make `lastScanOffset` non-optional and just set it to `0` whenever we don't have one.

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,91 @@
...
+        if let frame = try self.findNextFrame(buffer: &buffer) {
+            ctx.fireChannelRead(wrapInboundOut(frame))
+            return .continue
+        } else {
+            return .needMoreData
+        }
+    }
+    
+    private func findNextFrame(buffer: inout ByteBuffer) throws -> ByteBuffer? {
+        let view = buffer.readableBytesView.dropFirst(lastScanOffset)
```
Lukasa: Nit: `self` on `lastScanOffset`.

Sources/NIOExtras/LineBasedFrameDecoder.swift
```diff
@@ -0,0 +1,91 @@
...
+        self.handleLeftOverBytes(ctx: ctx)
+    }
+    
+    public func channelInactive(ctx: ChannelHandlerContext) {
+        self.handleLeftOverBytes(ctx: ctx)
+    }
+    
+    private func handleLeftOverBytes(ctx: ChannelHandlerContext) {
+        if let buffer = cumulationBuffer, buffer.readableBytes > 0 {
+            ctx.fireErrorCaught(NIOExtrasErrors.LeftOverBytesError(leftOverBytes: buffer))
```
weissi: - before calling out, you'll need to `self.cumulationBuffer.clear()` otherwise you'll deliver that twice
- also: explicit `self.`

Tests/NIOExtrasTests/LineBasedFramceDecoderTest.swift
```diff
@@ -0,0 +1,155 @@
...
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import NIO
+import NIOExtras
+
+class LineBasedFrameDecoderTest: XCTestCase {
+    
+    func testDecodeOneCharacterAtATime() throws {
+        let channel = EmbeddedChannel()
```
weissi: the setup (create channel & handler, add to pipeline) and tear down (finish) looks always the same, should we move that to a `func setup()`/`func tearDown`?

Sources/NIOExtras/LineBasedFrameDecoder.swift
ldewailly: Done. Added an assertion in the tests too to make sure the buffer has been cleared

PR #15, opened 2018-10-16
Add operators to `EventLoopFuture`

MrLotU:
Adds all possible operators from `Numeric` and `BinaryInteger` to `EventLoopFuture` + some

### Motivation:

Adding the ability to perform most "standard" operations between two EventLoopFuture's directly, without having to `.map` them first

### Modifications:

Added 33 operators to `EventLoopFuture`
### Result:

No breaking changes, but allows for things like this:
```swift
var varOne: EventLoopFuture<Int> = ...
var varTwo: EventLoopFuture<Int> = ...

var addedUp = varOne + varTwo // addedUp: EventLoopFuture<Int>
```


calebkleveter:
@Lukasa Does NIO core mean any NIO package maintained by the core NIO team? If so, what sort of functionality would go into the `swift-nio-extras` repo?

MrLotU:
@calebkleveter What and where? 😅

Lukasa:
@calebkleveter It’s a vague one, for sure, but I have mental gradations of what counts as “core”. In this case it’s not as core as the swift-nio repository, but it remains a place where there is code that requires maintenance and support. If the requirement is not something fairly widely needed, I’m inclined to want to let someone who is using the functionality maintain it, rather than have the NIO team take that responsibility. 

However, as noted above, this is an issue where I’m quite happy to have the rest of the team disagree with me. I don’t feel strongly here, because of how in the grey area it is.

MrLotU:
@Lukasa I see where you're comming from, although unless the nio teams decides to completely change the way Futures work, or the way numbers work in Swift, I don't think this code will require much "maintenance". If the rest of the team agrees with you though, I'm happy to host it on my own

Lukasa:
All code requires maintenance. 😁 In this case bear in mind that this could be considered a thin end of a wedge: why not provide similar conformances for Equatable, or Comparable, or indeed any other protocol that provides functions solely of the form `static func foo<T: SomeProtocol, U>(_ a: T, _ b: T) -> U`? So the total maintenance burden is, in fact, the hypothetical space of reasonable protocols to support. 

Which raises the question: would it not be better to solve the general form of this? We have `EventLoopFuture.reduce` in NIO today. We could add `EventLoopFuture.combine(with:using:)` to turn all of these functions into trivial one-liners, which I’d argue is more valuable than actually providing the conformance. 

Incidentally, is there a reason these shouldn’t be considered trivial

MrLotU:
Ah, yeah, in that case it would require quite some maintenance.

As for the 2nd part of your comment, If you guys think that's a better approach, I can look in getting something like that working too :)

MrLotU:
Because of @Lukasa's previous remarks about this preferrable not being in a NIO core repository. I've decided to move the PR to the new vapor-community/nio-kit repository. You can see the PR here: https://github.com/vapor-community/nio-kit/pull/3

Lukasa:
Sounds like a solid plan! 👍

PR #17, opened 2018-11-08
Adds a basic length field based frame decoder class.

### LINKED ISSUE #18: Length-delimited frame decoder.
It would be useful to have a general `ChannelInboundHandler` that can decode a length-delimited framing protocol. This is a protocol that has messages prefixed with an encoding of their length.

Such a protocol would look like this:

```
<fixed width integer n in network byte order><n bytes>
```

The simplest version of this would be one that uses a fixed-width integer.

For now we won't bother with varint support.

tigerpixel:
Adds a basic length field based frame decoder class.

### Motivation:
Adding a popular type of decoder that is useful in real-world situations, particularly when dealing with protocol buffers.

### Modifications:
Added the decoder class, tests and Linux test files.

### Result:
- The project now includes a basic length field based decoder which can be built upon.
- resolves #18


Tests/NIOExtrasTests/LengthFieldBasedFrameDecoderTest.swift
```diff
@@ -0,0 +1,298 @@
...
+        buffer.write(string: standardDataString)
+        
+        XCTAssertTrue(try channel.writeInbound(buffer))
+        
+        var outputBuffer: ByteBuffer? = channel.readInbound()
+        let outputData = outputBuffer?.readString(length: standardDataString.count)
+        
+        XCTAssertEqual(standardDataString, outputData)
+        XCTAssertFalse(try channel.finish())
+    }
```
Lukasa: I don't think this test is meaningfully different from the one above, we can remove it.

Tests/NIOExtrasTests/LengthFieldBasedFrameDecoderTest.swift
```diff
@@ -0,0 +1,298 @@
...
+    
+    func testDecodeWithUInt16HeaderWithString() throws {
+        
+        decoderUnderTest = LengthFieldBasedFrameDecoder(lengthFieldLength: 2)
+        try? channel.pipeline.add(handler: decoderUnderTest).wait()
+        
+        var dataLength: UInt16 = 5
+        
+        let headerData = Data(bytes: &dataLength, count: MemoryLayout.size(ofValue: dataLength))
```
Lukasa: This test is very subtly wrong if run on platforms that are not little-endian. This is because the default behaviour of the `LengthFieldBasedFrameDecoder` is to decode integers in a little-endian representation, and if the platform in question is big-endian then the integer will be incorrectly decoded.

Rather than use `Data` as a temporary intermediary, why not use `ByteBuffer.write(integer:)` to write the `UInt16`? That will allow you to ensure the the write uses the same endianness as the `LengthFieldBasedFrameDecoder`.

Incidentally, there are no tests that exercise the endianness parameter: perhaps there should be.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,122 @@
...
+
+    ///
+    /// Decodes the specified region of the buffer into an unadjusted frame length. The default implementation is
+    /// capable of decoding the specified region into an unsigned 8/16/32/64 bit integer.
+    /// - parameters:
+    ///    - buffer: The buffer containing the integer frame length
+    ///    - length: The length of the integer contained within the buffer.
+    ///    - endianness: The endianness of the integer contained within the buffer.
+    ///
+    private func frameLength(for buffer: ByteBuffer, length: Int, endianness: Endianness) -> Int? {
```
Lukasa: `endianness` doesn't need to be a parameter to this function, it's stored on the class.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,122 @@
...
+    private let lengthFieldLength: Int
+    private let lengthFieldEndianness: Endianness
+    
+    /// Create `LengthFieldBasedFrameDecoder` with a given frame length.
+    ///
+    /// - parameters:
+    ///    - lengthFieldLength: The length of the field specifying the remaining length of the frame.
+    ///    - lengthFieldEndianness: The endianness of the field specifying the remaining length of the frame.
+    ///
+    public init(lengthFieldLength: Int, lengthFieldEndianness: Endianness = .little) {
```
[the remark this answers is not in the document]
Lukasa: This is not a great choice for the default endianness. In general there are three major byte orders: little endian, big endian, and "network byte order". "Network byte order" is so-called because it is the byte ordering used by almost all network protocols, and it is the same as big endian.

For this purpose, we should default to big endian: users should be required to opt-in to little-endian.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,122 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+///
+/// A decoder that splits the received `ByteBuffer` by the number of bytes speicifed in a fixed length header
```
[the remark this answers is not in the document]
tigerpixel: will change

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,122 @@
...
+/// a `LengthFieldBasedFrameDecoder` will decode them into the following packets:
+///
+///     +-----+------+
+///     | BCD | FGHI |
+///     +-----+------+
+///
+/// 'A' and 'E' will be the headers and will not be passed forward.
+///
+
+public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
```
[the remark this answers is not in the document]
tigerpixel: I have used this class in Netty, where it has this name. I didn't really consider it too much because of that. I was planning to send future pull requests to make a class where there is a fixed length header and the data length specification can be anywhere in that header. That said, I think I prefer the use of 'prefix' and it would still fit. I am happy for you to make a call, it probably matters more to the project that the individual class. Also, maybe "LengthHeaderFrameDecoder".

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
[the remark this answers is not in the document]
tigerpixel: Okay, will do both of those.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,122 @@
...
+    }
+    
+    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
+        
+        guard let lengthFieldSlice = buffer.readSlice(length: lengthFieldLength) else {
+            return .needMoreData
+        }
+
+        // convert the length field to an int specifying the length
+        guard let lengthFieldValue = frameLength(for: lengthFieldSlice,
```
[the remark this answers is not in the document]
tigerpixel: Sure. Can I ask, for future reference? Is there a style doc I should be looking at, other than the Swift language style guide.

Tests/NIOExtrasTests/LengthFieldBasedFrameDecoderTest.swift
tigerpixel: will remove

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: will remove

Tests/NIOExtrasTests/LengthFieldBasedFrameDecoderTest.swift
tigerpixel: Sure, all good points. Will change.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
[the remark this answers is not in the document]
Lukasa: Were you planning to make a *new* class that did this, or extend this one? If you're extending this one, I'm happy for this one to have the name `LengthFieldBasedFrameDecoder`. Otherwise, I think we should change the name of this class, as `LengthFieldBasedFrameDecoder` in `Netty` is the class that lets you select where the length field is in the header.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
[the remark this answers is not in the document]
Lukasa: I'm afraid we haven't written one yet: right now it is entirely tribal knowledge. 😞

tigerpixel:
Hi, just to let you know. I have either changed or answered all outstanding. I am happy to make further changes if you think they are required. Thanks.

carlos4242:
p.s. ... https://forums.swift.org/t/happy-thanksgiving-a-k-a-where-are-all-the-apple-folks/18070

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,113 @@
...
+
+///
+/// An enumeration to describe the length of a piece of data in bytes.
+/// It is contained to lengths that can be converted to integer types.
+///
+public enum ByteLength: Int {
+    case one = 1
+    case two = 2
+    case four = 4
+    case eight = 8
```
Lukasa: I don't think this `enum` needs a raw value.

Tests/NIOExtrasTests/LengthFieldBasedFrameDecoderTest.swift
```diff
@@ -0,0 +1,297 @@
...
+        XCTAssertFalse(try self.channel.finish())
+    }
+    
+    func testEmptyBuffer() throws {
+        
+        self.decoderUnderTest = LengthFieldBasedFrameDecoder(lengthFieldLength: .one,  lengthFieldEndianness: .little)
+        try? self.channel.pipeline.add(handler: self.decoderUnderTest).wait()
+        
+        var buffer = self.channel.allocator.buffer(capacity: 1)
+        buffer.write(string: "")
```
Lukasa: We don't need this `write`.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: I just moved it to a fileprivate method as it is likely to be used more than once if the class evolves.

Lukasa:

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,130 @@
...
+        guard let lengthFieldSlice = buffer.readSlice(length: self.lengthFieldLength.length) else {
+            return .needMoreData
+        }
+
+        // convert the length field to an int specifying the length
+        guard let lengthFieldValue = self.frameLength(for: lengthFieldSlice) else {
+            return .needMoreData
+        }
+
+        guard let contentsFieldSlice = buffer.readSlice(length: lengthFieldValue) else {
```
weissi: I don’t think this actually works. Let’s assume we get the following bytes (and length `.one`):

`5, 1, 2, 3`

Then nothing for a while, and then

`4, 5`

What we would expect is that the first `5` is the length (5 bytes) and the next 1, 2, 3, 4, 5 are the data.

With your code however, we’ll find the first 5 to be the length (correct), then we’ll try to read 5 bytes (correct but unavailable).
As soon as more bytes arrive, we’ll start the parsing loop again and we’ll interpret the `4` and another length field (incorrect).

I think you’ll want a inner type

```
enum State {
    case waitingForHeader
    case waitingForData(length: Int)
}
```

And `var state: State` and depending on the state you do the right thing in the parsing loop.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,130 @@
...
+
+        switch self.lengthFieldLength {
+        case .one:
+            return buffer.getInteger(at: 0, endianness: self.lengthFieldEndianness, as: UInt8.self).map { Int($0) }
+        case .two:
+            return buffer.getInteger(at: 0, endianness: self.lengthFieldEndianness, as: UInt16.self).map { Int($0) }
+        case .four:
+            return buffer.getInteger(at: 0, endianness: self.lengthFieldEndianness, as: UInt32.self).map { Int($0) }
+        case .eight:
+            return buffer.getInteger(at: 0, endianness: self.lengthFieldEndianness, as: UInt64.self).map { Int($0) }
```
weissi: This will fail on 32 bit platforms. You’ll need to use `UInt64` to parse the length information.

Lukasa: @weissi It's a bit more awkward than that. This value is being passed to `readSlice`, so it's possible that on 32-bit platforms we'll be asked to read more from the slice than is possible to represent in the 32-bit `Int`. More importantly, `readSlice` doesn't take a `UInt64`, it takes an `Int`, so whatever is returned here has to be converted into an `Int` anyway.

Frankly, I think the edge-case isn't worth worrying about: if a 32-bit machine is asked to read more than `Int.max` from a frame something bad is about to happen anyway.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: Looks like a good point. Either way it needs a test. So, I will add that first and then fix the problem once confirmed.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: What about if we guard against the use of 64bit length field on a 32bit platform in the constructor? I can add a comment next to it and it saves anyone getting into a pickle but gets this through without a major refactor. I am happy to do the work if required, it just sounds like it is slightly out of the scope of this PR.

weissi: @tigerpixel / @Lukasa I think we should guard in the constructor, otherwise this is a good way to mount a DOS attack. The attacker can just send `0xffffffff` as the length to a 32 bit system and frameLength will always return `nil` so it'll always return `.needMoreData` which will cause the `ByteToMessageDecoder` to read arbitrary amounts of data into memory.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: Code changes were needed. Good spot.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: Added.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -42,6 +42,15 @@ extension ByteLength {
     }
 }
 
+///
+/// The state of a decoder. It has two distinct sections of data to read. Each must be fully present before it is considered as read.
+/// During the time when it is not present the decoder must wait. `DecoderReadState` details that waiting state.
+///
+private enum DecoderReadState {
```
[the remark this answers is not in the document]
tigerpixel: Good point. Is neater. Done.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,170 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+///
+/// An enumeration to describe the length of a piece of data in bytes.
+/// It is contained to lengths that can be converted to integer types.
+///
+public enum ByteLength {
```
weissi: given that this is a public type, mind putting this inside of your decoder? Otherwise it pollutes the `NIOExtras` namespace with this type. Anyway I think `LengthFieldBasedFrameDecoder.ByteLength` is a better name

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,170 @@
...
+        return .continue
+    }
+    
+    private func readNextLengthFieldToState(buffer: inout ByteBuffer) throws {
+        
+        guard let lengthFieldSlice = buffer.readSlice(length: self.lengthFieldLength.length) else {
+            return
+        }
+        
+        // Convert the length field to an integer specifying the length
```
weissi: this still can't be right. You're mutating `buffer` (by reading something off) but you're not mutating `self.readState`.

Generally, `self.readState` will need to remember all changes you have already done where there's a way to return in the mean time. So now, you'll have an issue (and this needs a test case) if you receive the bytes in this order:

- `channelRead`: just a `5`
- now we read off a slice `lengthFieldSlice` (correct)
- then we try to do `self.frameLength` but if that fails, we'll return having read `lengthFieldSlice` off already but we'll still be in the `.waitingForHeader` state (incorrect)

So if we get bytes again, say `1, 2, 3, 4, 5` (which should be the data), we'll interpret the `1` as a new header...

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,170 @@
...
+        }
+        
+        ctx.fireChannelRead(self.wrapInboundOut(frameBuffer))
+
+        return .continue
+    }
+    
+    private func readNextLengthFieldToState(buffer: inout ByteBuffer) throws {
+        
+        guard let lengthFieldSlice = buffer.readSlice(length: self.lengthFieldLength.length) else {
```
weissi: we're reading a `lengthFieldSlice` off here but we're never actually using it. I think you should return `Int?` here and just read off the `headerField`. Basically the same as what you have in `func frameLength` but just use `readInteger` instead of `getInteger` there.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    ///
+    /// Decodes the specified region of the buffer into an unadjusted frame length. The default implementation is
+    /// capable of decoding the specified region into an unsigned 8/16/32/64 bit integer.
+    /// - parameters:
+    ///    - buffer: The buffer containing the integer frame length
+    ///
+    private func frameLength(for buffer: ByteBuffer) -> Int? {
```
weissi: regarding my comment above: I think the fix is really easy:
- just rename this function from `frameLength` to `readFrameLength`
- change all `getInteger` in there to `readInteger`
- remove the whole `guard let lengthFieldSlice = buffer.readSlice(length: self.lengthFieldLength.length) else { [...] }` block above and I think then it'll work already :)

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: Okay, so I am going to write a test where the data come in and is read with single bytes, so that if there is an error by either splitting the header or the data it will catch it.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: Yea, good point. I think that 2 stage check is a product of a few small refactorings and should probably go anyway, even discounting the fact that you have found a bug in it. Will fix.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: Agreed.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
weissi: sounds great, thank you!

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: Sure, done.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
tigerpixel: So, I split the test into single bytes. It didn't catch this error. I think because it is actually very difficult to get the first call to succeed and the second to fail. As mentioned I neatened it out anyway as it is better not to have the call twice.

weissi:
@swift-nio-bot please test

weissi:

PR #19, opened 2018-11-30
Adds a LengthFieldPrepender class to prepend the length onto a message.

tigerpixel:
Adds a type of byte to message encoder to prepend the length onto a message.

Motivation:
To encode a prepended length field on data so that messages of arbitrary size can be sent.
Can work as a pair with the ‘LengthFieldBasedFrameDecoder’.

Modifications:
Added ‘LengthFieldPrepender’
Added unit tests for ‘LengthFieldPrepender’ in ‘LengthFieldPrependerTest’
Updated the linux text files by running the script.

Result:
The length can now be easily prepended to any message.


Sources/NIOExtras/LengthFieldPrepender.swift
```diff
@@ -0,0 +1,97 @@
...
+            out.write(integer: UInt8(data.readableBytes), endianness: self.lengthFieldEndianness)
+        case .two:
+            out.write(integer: UInt16(data.readableBytes), endianness: self.lengthFieldEndianness)
+        case .four:
+            out.write(integer: UInt32(data.readableBytes), endianness: self.lengthFieldEndianness)
+        case .eight:
+            out.write(integer: UInt64(data.readableBytes), endianness: self.lengthFieldEndianness)
+        }
+        
+        out.write(bytes: data.readableBytesView)
```
tigerpixel: This passes the unit tests and works in my little test program, but I am not sure about the efficiency of writing using the byte view. If there is a more efficient way, would you mind advising so that I can take a look at a preferable way to achieve the same goal. Thanks.

weissi: `out.write(buffer: data)` would be better here. However that will still do a copy that we don’t need.
If you make the whole class a `ChannelOutboundHandler` and in `func write` you first write the length and then just forward the incoming write, you can do this whole thing without allocating and without copying and that’ll be much more efficient.

NIO is smart enough that if you do

```
ctx.write(self.wrapOutboundOut(self.lengthBuffer), promise: nil)
ctx.write(data, promise: promise)
```

Both these writes will happen in one system call :)

tigerpixel: Perfect, exactly what I was looking for. Thanks. Will do.

tigerpixel: Okay, all done. Thanks for the tip.

Sources/NIOExtras/LengthFieldPrepender.swift
```diff
@@ -0,0 +1,122 @@
...
+/// An encoder that takes a `ByteBuffer` message and prepends the number of bytes in the message.
+/// The length field is always the same fixed length specified on construction.
+/// These bytes contain a binary specification of the message size.
+///
+/// For example, if you received a packet with the 3 byte length (BCD)...
+/// Given that the specified header length is 1 byte, there would be a single byte appended which contains the number 3
+///     +---+-----+
+///     | A | BCD | ('A' contains 0x0003)
+///     +---+-----+
+/// This initial appended byte is called the 'length field'.
```
Lukasa: Nit: not appended, prepended.

Sources/NIOExtras/LengthFieldPrepender.swift
```diff
@@ -0,0 +1,122 @@
...
+///
+public final class LengthFieldPrepender: ChannelOutboundHandler {
+    
+    ///
+    /// An enumeration to describe the length of a piece of data in bytes.
+    /// It is contained to lengths that can be converted to integer types.
```
Lukasa: Nit: not contained, constrained.

Sources/NIOExtras/LengthFieldPrepender.swift
```diff
@@ -0,0 +1,122 @@
...
+
+        let dataBuffer = self.unwrapOutboundIn(data)
+        let dataLength = dataBuffer.readableBytes
+        
+        guard dataLength <= self.lengthFieldLength.max else {
+            promise?.fail(error: LengthFieldPrependerError.messageDataTooLongForLengthField)
+            return
+        }
+
+        var lengthBuffer = ctx.channel.allocator.buffer(capacity: self.lengthFieldLength.length)
```
Lukasa: Rather than allocate each time, should we pre-allocate this buffer in `handlerAdded` and then simply re-use it via `clear`?

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+        
+        try self.channel.writeAndFlush(buffer).wait()
+        
+        if case .some(.byteBuffer(var headerBuffer)) = self.channel.readOutbound() {
+            
+            let outputData = headerBuffer.readBytes(length: headerBuffer.readableBytes)
+            XCTAssertEqual([UInt8(dataBytes.count)],  outputData)
+            
+            let additionalData = headerBuffer.readBytes(length: 1)
+            XCTAssertNil(additionalData)
```
Lukasa: This assertion is always true by the definition of `readBytes` with a length of `readableBytes`. This will consume all of the data in the buffer, so this subsequent read must fail. We may as well remove it.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+            XCTFail("couldn't read ByteBuffer from channel")
+        }
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let outputData = outputBuffer.readBytes(length: outputBuffer.readableBytes)
+            XCTAssertEqual(dataBytes,  outputData)
+            
+            let additionalData = outputBuffer.readBytes(length: 1)
+            XCTAssertNil(additionalData)
```
Lukasa: Same with this assertion.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+    func testEncodeWithUInt16HeaderWithString() throws {
+        
+        let endianness: Endianness = .little
+        
+        self.encoderUnderTest = LengthFieldPrepender(lengthFieldLength: .two,
+                                                     lengthFieldEndianness: endianness)
+        
+        try? self.channel.pipeline.add(handler: self.encoderUnderTest).wait()
+        
+        var buffer = self.channel.allocator.buffer(capacity: standardDataString.count)
```
Lukasa: This use of `.count` is not really safe, and it only works because we have an ASCII string. It'd be better to use `standardDataString.utf8`.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+        buffer.write(string: standardDataString)
+        
+        try self.channel.writeAndFlush(buffer).wait()
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let sizeInHeader = outputBuffer.readInteger(endianness: endianness, as: UInt16.self).map({ Int($0) })
+            XCTAssertEqual(standardDataString.count, sizeInHeader)
```
Lukasa: This use of count is also not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+            let additionalData = outputBuffer.readBytes(length: 1)
+            XCTAssertNil(additionalData)
+            
+        } else {
+            XCTFail("couldn't read ByteBuffer from channel")
+        }
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let bodyString = outputBuffer.readString(length: standardDataString.count)
```
Lukasa: This use of count is also not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+    func testEncodeWithUInt32HeaderWithString() throws {
+        
+        let endianness: Endianness = .little
+        
+        self.encoderUnderTest = LengthFieldPrepender(lengthFieldLength: .four,
+                                                     lengthFieldEndianness: endianness)
+        
+        try? self.channel.pipeline.add(handler: self.encoderUnderTest).wait()
+        
+        var buffer = self.channel.allocator.buffer(capacity: standardDataString.count)
```
Lukasa: This use of `.count` is also not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+            let additionalData = outputBuffer.readBytes(length: 1)
+            XCTAssertNil(additionalData)
+            
+        } else {
+            XCTFail("couldn't read ByteBuffer from channel")
+        }
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let bodyString = outputBuffer.readString(length: standardDataString.count)
```
Lukasa: This use of `.count` isn't safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+    func testEncodeWithUInt64HeaderWithString() throws {
+        
+        let endianness: Endianness = .little
+        
+        self.encoderUnderTest = LengthFieldPrepender(lengthFieldLength: .eight,
+                                                     lengthFieldEndianness: endianness)
+        
+        try? self.channel.pipeline.add(handler: self.encoderUnderTest).wait()
+        
+        var buffer = self.channel.allocator.buffer(capacity: standardDataString.count)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+        buffer.write(string: standardDataString)
+        
+        try self.channel.writeAndFlush(buffer).wait()
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let sizeInHeader = outputBuffer.readInteger(endianness: endianness, as: UInt64.self).map({ Int($0) })
+            XCTAssertEqual(standardDataString.count, sizeInHeader)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+            let additionalData = outputBuffer.readBytes(length: 1)
+            XCTAssertNil(additionalData)
+            
+        } else {
+            XCTFail("couldn't read ByteBuffer from channel")
+        }
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let bodyString = outputBuffer.readString(length: standardDataString.count)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+    func testEncodeWithInt64HeaderWithString() throws {
+        
+        let endianness: Endianness = .little
+        
+        self.encoderUnderTest = LengthFieldPrepender(lengthFieldLength: .eight,
+                                                     lengthFieldEndianness: endianness)
+        
+        try? self.channel.pipeline.add(handler: self.encoderUnderTest).wait()
+        
+        var buffer = self.channel.allocator.buffer(capacity: standardDataString.count)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+        buffer.write(string: standardDataString)
+        
+        try self.channel.writeAndFlush(buffer).wait()
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let sizeInHeader = outputBuffer.readInteger(endianness: endianness, as: Int64.self).map({ Int($0) })
+            XCTAssertEqual(standardDataString.count, sizeInHeader)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+    func testEncodeWithUInt64HeaderStringBigEndian() throws {
+        
+        let endianness: Endianness = .big
+        
+        self.encoderUnderTest = LengthFieldPrepender(lengthFieldLength: .eight,
+                                                     lengthFieldEndianness: endianness)
+        
+        try? self.channel.pipeline.add(handler: self.encoderUnderTest).wait()
+        
+        var buffer = self.channel.allocator.buffer(capacity: standardDataString.count)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+        buffer.write(string: standardDataString)
+        
+        try self.channel.writeAndFlush(buffer).wait()
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let sizeInHeader = outputBuffer.readInteger(endianness: endianness, as: UInt64.self).map({ Int($0) })
+            XCTAssertEqual(standardDataString.count, sizeInHeader)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+            let additionalData = outputBuffer.readBytes(length: 1)
+            XCTAssertNil(additionalData)
+            
+        } else {
+            XCTFail("couldn't read ByteBuffer from channel")
+        }
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+
+            let bodyString = outputBuffer.readString(length: standardDataString.count)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+        XCTAssertFalse(try self.channel.finish())
+    }
+    
+    func testEncodeWithInt64HeaderStringDefaultingToBigEndian() throws {
+        
+        self.encoderUnderTest = LengthFieldPrepender(lengthFieldLength: .eight)
+        
+        try? self.channel.pipeline.add(handler: self.encoderUnderTest).wait()
+        
+        var buffer = self.channel.allocator.buffer(capacity: standardDataString.count)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+        buffer.write(string: standardDataString)
+        
+        try self.channel.writeAndFlush(buffer).wait()
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let sizeInHeader = outputBuffer.readInteger(endianness: .big, as: UInt64.self).map({ Int($0) })
+            XCTAssertEqual(standardDataString.count, sizeInHeader)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+            let additionalData = outputBuffer.readBytes(length: 1)
+            XCTAssertNil(additionalData)
+            
+        } else {
+            XCTFail("couldn't read ByteBuffer from channel")
+        }
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let bodyString = outputBuffer.readString(length: standardDataString.count)
```
Lukasa: This use of count is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+            XCTFail("couldn't read ByteBuffer from channel")
+        }
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+
+            let bodyData = outputBuffer.readBytes(length: contents.count)
+            XCTAssertEqual(contents, bodyData)
+            
+            let additionalData = outputBuffer.readBytes(length: 1)
+            XCTAssertNil(additionalData)
```
Lukasa: Remove this assertion.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+            
+        } else {
+            XCTFail("couldn't read ByteBuffer from channel")
+        }
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+
+            let bodyData = outputBuffer.readBytes(length: contents.count)
```
Lukasa: Let's read the whole thing.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+        
+        var buffer = self.channel.allocator.buffer(capacity: standardDataString.count)
+        buffer.write(string: standardDataString)
+        
+        try self.channel.writeAndFlush(buffer).wait()
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+            
+            let sizeInHeader = outputBuffer.readInteger(endianness: endianness, as: UInt32.self).map({ Int($0) })
+            XCTAssertEqual(standardDataString.count, sizeInHeader)
```
Lukasa: This use of `.count` is not safe.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,424 @@
...
+            let additionalData = outputBuffer.readBytes(length: 1)
+            XCTAssertNil(additionalData)
+            
+        } else {
+            XCTFail("couldn't read ByteBuffer from channel")
+        }
+        
+        if case .some(.byteBuffer(var outputBuffer)) = self.channel.readOutbound() {
+
+            let bodyString = outputBuffer.readString(length: standardDataString.count)
```
Lukasa: This use of `.count` is not safe.

tigerpixel:
Hey, those should all be done now. Let me know if you have any issues.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,413 @@
...
+    override func setUp() {
+        self.channel = EmbeddedChannel()
+    }
+
+    func testEncodeWithUInt8HeaderWithData() throws {
+        
+        self.encoderUnderTest = LengthFieldPrepender(lengthFieldLength: .one,
+                                                     lengthFieldEndianness: .little)
+
+        try? self.channel.pipeline.add(handler: self.encoderUnderTest).wait()
```
weissi: nit, this should be

    XCTAssertNoThrow(try self.channel.pipeline.add(.....).wait()

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
```diff
@@ -0,0 +1,413 @@
...
+        
+        let dataBytes: [UInt8] = [10, 20, 30, 40]
+        
+        var buffer = self.channel.allocator.buffer(capacity: dataBytes.count)
+        buffer.write(bytes: dataBytes)
+        
+        try self.channel.writeAndFlush(buffer).wait()
```
weissi: nit: this should also be wrapped in an `XCTAssertNoThrow`, similarly all the others that throw.

Sources/NIOExtras/LengthFieldPrepender.swift
```diff
@@ -0,0 +1,132 @@
...
+
+///
+/// An encoder that takes a `ByteBuffer` message and prepends the number of bytes in the message.
+/// The length field is always the same fixed length specified on construction.
+/// These bytes contain a binary specification of the message size.
+///
+/// For example, if you received a packet with the 3 byte length (BCD)...
+/// Given that the specified header length is 1 byte, there would be a single byte prepended which contains the number 3
+///     +---+-----+
+///     | A | BCD | ('A' contains 0x0003)
```
Lukasa: Nit: should be `0x03`, `0x0003` is two bytes.

Tests/NIOExtrasTests/LengthFieldPrependerTest.swift
tigerpixel: Added.

Sources/NIOExtras/LengthFieldPrepender.swift
tigerpixel: Haha, true, the perils of changing an example at the last minute. Thanks.

tigerpixel:
Okay, nits hopefully addressed.

weissi:
thank you @tigerpixel !

weissi:

weissi:
@Lukasa ping

PR #20, opened 2018-12-07
LengthFieldBasedFrameDecoder: work around brittle B2MD

weissi:
Motivation:

ByteToMessageDecoder is extremely brittle, for example a reentrant call
into decodeLast will present the user with bytes that were previously
seen...

Modification:

Discard bytes in decodeLast

Result:

LengthFieldBasedFrameDecoder will work if close called from channelRead.


weissi:
thanks done @Lukasa

weissi:
ping @Lukasa

PR #25, opened 2019-02-21
Swift 5 docker-compose stuff

weissi:


weissi:
@tomerd thanks, I fixed that and just ran the docker-compose which ran through ok

PR #27, opened 2019-02-26
fix docker setup for swift 5

tomerd:
motivation: build on swift 5

changes:
* fix test task defition to use correct image
* remove redundant jazzy setup


tomerd:
@weissi

PR #46, opened 2019-04-02
write PCAP handler

weissi:
Motivation:

Especially with TLS, but also for example if you don't have the privileges to dump network traffic using `tcpdump` in production, it can be handy to be able to write `.pcap` files from NIO directly. With this PR, SwiftNIO synthesises TCP over IPv4/6 packets as if they were real packets so you can inspect them in Wireshark/`tcpdump -r`. This is also super useful for Unix Domain Sockets because there are no sniffing/dumping tools, in the Unix Domain Socket case we'll just use fake IPv4 addresses. With `NIOWritePCAPHandler` that you can insert into your `ChannelPipeline` where ever you send bytes around.

Modifications:

add a ChannelHandler that can write a PCAP trace from what's going on in
the ChannelPipeline.

Result:

easier debugging in production


weissi:
Screenshot 2019-04-02 at 10 13 17 pm

SUPER WIP, but wireshark kind of understands what's going on already :)

weissi:
Screenshot 2019-04-03 at 6 38 48 pm

it actually works properly now. Also supports IPv6 now and even Unix Domain Sockets (using fake IP addresses :) )

weissi:
@Lukasa / @tomerd  I would love to test this but the best testing strategy (short of writing a PCAP parser (which I won't be doing 😬)) is to also add the [bash integration test framework](https://github.com/apple/swift-nio/tree/master/IntegrationTests) from `swift-nio` to this repo and just running `tcpdump -r FILE -XXX` and then testing certain things...

Any better ideas?

Lukasa:
I think that idea is fine tbh.

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -0,0 +1,649 @@
...
+        self.mode = mode
+        if let fakeLocalAddress = fakeLocalAddress {
+            self.localAddress = fakeLocalAddress
+        }
+        if let fakeRemoteAddress = fakeRemoteAddress {
+            self.remoteAddress = fakeRemoteAddress
+        }
+    }
+    
+    private func writeBuffer(_ buffer: ByteBuffer, context: ChannelHandlerContext) {
```
Lukasa: This doesn't seem to need the `ChannelHandlerContext` anymore.

weissi: done

PR #48, opened 2019-04-12
Add a podspec for NIOExtras

### LINKED ISSUE #45: make cocoapod
NIOExtras needs a cocoapod

Palleas:
Motivation:

https://github.com/apple/swift-nio-extras/issues/45

Modifications:

Add a podspec, you should be able to validate it with `pod lib lint` and `pod spec lint`.

Result:

Add podspec will be available to the users. Might need to be pushed
somewhere though.

Closes #45.

### DIFF
--- a/scripts/build_podspec.sh
+++ b/scripts/build_podspec.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2017-2019 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+set -eu
+
+function usage() {
+  echo "$0 [-u] version"
+  echo
+  echo "OPTIONS:"
+  echo "  -u: Additionally upload the podspec"
+}
+
+upload=false
+while getopts ":u" opt; do
+  case $opt in
+    u)
+      upload=true
+      ;;
+    \?)
+      usage
+      exit 1
+      ;;
+  esac
+done
+shift "$((OPTIND-1))"
+
+if [[ $# -eq 0 ]]; then
+  echo "Must provide target version"
+  exit 1
+fi
+
+version=$1
+name="SwiftNIOExtras"
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+tmpdir=$(mktemp -d /tmp/.build_podspecsXXXXXX)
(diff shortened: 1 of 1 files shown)


Lukasa:
Hey @Palleas! I'm happy to accept a podspec for swift-nio-extras, but for ease of maintenance it'd be nicer if you could refactor this a bit so that we have a podspec-generating script, rather than a podspec itself. We use this in `swift-nio-transport-services` to make it easier for us to ship new releases, as you can [see here](https://github.com/apple/swift-nio-transport-services/blob/master/scripts/build_podspec.sh).

weissi:
@Palleas thanks! Here's another one I just proposed for `swift-log`: https://github.com/apple/swift-log/pull/54

Palleas:
Oops, I should have checked the other projects 🙈 I'll fix that today!

Palleas:
@weissi @Lukasa done! Here is the output:


```ruby
Pod::Spec.new do |s|
  s.name = 'NIOExtras'
  s.version = '1.0.0'
  s.license = { :type => 'Apache 2.0', :file => 'LICENSE.txt' }
  s.summary = 'Useful code around SwiftNIO.'
  s.homepage = 'https://github.com/apple/swift-nio-extras'
  s.author = 'Apple Inc.'
  s.source = { :git => 'https://github.com/apple/swift-nio-extras.git', :tag => s.version.to_s }
  s.documentation_url = 'https://github.com/apple/swift-nio-extras'
  s.module_name = 'NIOExtras'

  s.swift_version = '5.0'
  s.cocoapods_version = '>=1.6.0'
  s.ios.deployment_target = '12.0'
  s.osx.deployment_target = '10.14'
  s.tvos.deployment_target = '12.0'

  s.source_files = 'Sources/NIOExtras/**/*.swift'
end
```

scripts/build_podspec.sh
```diff
@@ -0,0 +1,78 @@
...
+done
+shift "$((OPTIND-1))"
+
+if [[ $# -eq 0 ]]; then
+  echo "Must provide target version"
+  exit 1
+fi
+
+version=$1
+NAME="NIOExtras"
```
weissi: - this should be `SwiftNIOExtras` to match the others (reason is that `NIO` was taken) 
- `NAME` should be lower-case, all caps is reserved shell variables

scripts/build_podspec.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+
```
weissi: I'm fairly sure the sanity checking script will complain about the newline between the shebang and the license header. You can run it on your machine with `scripts/sanity.sh` and it should not output errors.

scripts/build_podspec.sh
Palleas: Done! I also added `SwiftNIO` as a dependency, otherwise it would not compile properly. 🤦‍♂️

scripts/build_podspec.sh
Palleas: Done!

weissi:

PR #52, opened 2019-05-10
use B2MD verifier

weissi:
Motivation:

Use B2MDVerifier for the B2MDs in NIOExtras. Already found one bug,
separetely fixed in #51.

Modifications:

Write a basic validation test for all B2MDs.

Result:

Better test coverage.


Tests/NIOExtrasTests/FixedLengthFrameDecoderTest.swift
```diff
@@ -115,30 +116,22 @@ class FixedLengthFrameDecoderTest: XCTestCase {
         XCTAssertTrue(try channel.finish().isClean)
     }
 
-    func testCloseInChannelRead() {
```
weissi: just in case anyone is surprised: this test was in the wrong test suite, I moved it.

PR #53, opened 2019-05-11
JSONRPC framing: Content-Length

weissi:
Motivation:

JSPN-RPC uses various framing methods, one is Content-Length based
framing. NIOExtras should provide encoder/decoders for this.

Modifications:

Add such codecs.

Result:

NIOExtras more useful


Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
```diff
@@ -0,0 +1,194 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+extension NIOJSONRPCFraming {
```
weissi: @tomerd I nested this in `NIOJSONRPCFraming` so you can add your codecs into this too so we don't pollute the name-space too much.

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
```diff
@@ -0,0 +1,194 @@
...
+            case .waitingForHeaderNameOrHeaderBlockEnd:
+                // Given that we're waiting for the end of a header block or a new header field, it's sensible to
+                // check if this might be the end of the block.
+                if buffer.readableBytesView.starts(with: "\r\n".utf8) {
+                    buffer.moveReaderIndex(forwardBy: 2) // skip \r\n\r\n
+                    if let payloadLength = self.payloadLength {
+                        if payloadLength == 0 {
+                            // special case, we're not actually waiting for anything if it's 0 bytes...
+                            self.state = .waitingForHeaderNameOrHeaderBlockEnd
+                            self.payloadLength = 0
```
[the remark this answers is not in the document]
Lukasa: Of course, the other option is to burn this branch away entirely: we'll do an extra spin through this state machine, but the behaviour will be right.

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
```diff
@@ -0,0 +1,194 @@
...
+    func trimmed() -> Substring {
+        var substring = self[...]
+        while substring.first.map({ $0.isNewline || $0.isWhitespace }) ?? false {
+            substring = substring.dropFirst()
+        }
+        while substring.last.map({ $0.isNewline || $0.isWhitespace }) ?? false {
+            substring = substring.dropLast()
+        }
+        return substring
+    }
```
Lukasa: Feel free to ignore the code golf, but this seems like a verbose spelling for:

```swift
    func trimmed() -> Substring {
        guard let firstElementIndex = self.firstIndex(where: { !$0.isWhitespace }) else {
            return Substring("")
        }

        let lastElementIndex = self.reversed().first(where: { !$0.isWhitespace })!
        return self[firstElementIndex...lastElementIndex]
    }
```

Either way, you can remove `isNewline` because `isWhitespace` covers it.

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
```diff
@@ -0,0 +1,194 @@
...
+        public init() {}
+
+        public func handlerAdded(context: ChannelHandlerContext) {
+            self.scratchBuffer = context.channel.allocator.buffer(capacity: 512)
+        }
+
+        public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+            let data = self.unwrapOutboundIn(data)
+            // Step 1, reserve enough space in the target buffer, this should mostly be a no-op
+            self.scratchBuffer.reserveCapacity(28 /* Content-Length: ______\r\n\r\n */)
```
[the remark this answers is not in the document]
weissi: @Lukasa thanks, indeed. The one in LSP was a bit different, didn't think when adapting.

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
```diff
@@ -0,0 +1,194 @@
...
+            if data.readableBytes > 0 {
+                context.write(self.wrapOutboundOut(self.scratchBuffer), promise: nil)
+                context.write(self.wrapOutboundOut(data), promise: promise)
+            } else {
+                context.write(self.wrapOutboundOut(self.scratchBuffer), promise: promise)
+            }
+        }
+    }
+
+    /// `ContentLengthHeaderFrameDecoder` is responsible for emitting JSON-RPC wire protocol with 'Content-Length'
```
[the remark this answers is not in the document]
weissi: close ;)

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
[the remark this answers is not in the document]
weissi: thanks, indeed!

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
```diff
@@ -0,0 +1,189 @@
...
+                        }
+                        // cool, let's just shift to the `.waitingForPayload` state and continue.
+                        self.state = .waitingForPayload(length: payloadLength)
+                        self.payloadLength = nil
+                        return .continue
+                    } else {
+                        // this happens if we reached the end of the header block but we haven't seen the Content-Length
+                        // header, that's an error. It will be sent through the `Channel` and decoder won't be called
+                        // again.
+                        throw DecodingError.missingContentLengthHeader
```
tomerd: nit: guard would remove some nesting

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
```diff
@@ -0,0 +1,189 @@
...
+                    }
+                }
+
+                // Given that this is not the end of a header block, it must be a new header field. A new header field
+                // must always have a colon (or we don't have enough data).
+                if let colonIndex = buffer.readableBytesView.firstIndex(of: UInt8(ascii: ":")) {
```
tomerd: would caching `UInt8(ascii: ":")` be more efficient?

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
```diff
@@ -0,0 +1,189 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+extension NIOJSONRPCFraming {
+    /// `ContentLengthHeaderFrameEncoder` is responsible for emitting JSON-RPC wire protocol with 'Content-Length'
+    /// HTTP-like headers as used by for example by LSP (Language Server Protocol).
+    public final class ContentLengthHeaderFrameEncoder: ChannelOutboundHandler {
```
tomerd: in some of the others we implemented max-size which could be important in real life use cases

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
weissi: @tomerd no, it would be less efficient. The compiler fully evaluates `UInt8(ascii: ":")` into whatever number it is.

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
weissi: @tomerd it's definitely relevant for the `Decoder` but we have max-size implemented in `ByteToMessageHandler` which means all `ByteToMessageDecoder`s can benefit from it. But this is the `Encoder` which will only be invoked if the _user_ already has that many bytes in memory. Therefore I don't think we need/should limit the maximum.

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
weissi: thanks @tomerd , done

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
tomerd: ah right, i forgot i implemented it there :)

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
tomerd: cool, til

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
weissi: haha right, you actually did :). Good job!

Sources/NIOExtras/JSONRPCFraming+ContentLengthHeader.swift
weissi: @tomerd just in case you want to try those things out yourself:

Step 1) Create a new swift file and put something like this in there:

```swift
@inline(never)
func SOMETHING_VERY_EASY_TO_GREP() -> Int {
    return Int(UInt8(ascii: "X"))
}

print(SOMETHING_VERY_EASY_TO_GREP())
```

then run this command:

```
$ swiftc -O -emit-assembly test.swift | grep -A10 SOMETHING_VERY_EASY
	.private_extern	_$s4test27SOMETHING_VERY_EASY_TO_GREPSiyF
	.globl	_$s4test27SOMETHING_VERY_EASY_TO_GREPSiyF
	.p2align	4, 0x90
_$s4test27SOMETHING_VERY_EASY_TO_GREPSiyF:
	pushq	%rbp
	movq	%rsp, %rbp
	movl	$88, %eax
	popq	%rbp
	retq

	.private_extern	_$ss23_ContiguousArrayStorageCyypGML
	.section	__DATA,__data
	.globl	_$ss23_ContiguousArrayStorageCyypGML
	.weak_definition	_$ss23_ContiguousArrayStorageCyypGML
```

and as you can see, the whole code of the function (modulo function start/end that would normally be inlined hadn't we done `@inline(never)` is

```
movl	$88, %eax
```

which just moves `88` into `eax`/`rax` which is the return register and 88 ascii is `X` 🙌

weissi: Summary: The compiler fully converts `Int(UInt(ascii: "X"))` to the constant `88` which is puts in where ever you use it.

ktoso: 👍 when feeling lazy you can also use (though less control about specific compiler version): https://godbolt.org

Screen Shot 2019-05-14 at 11 13 17

It's fun for exploring what gets generated.

weissi: @ktoso yes, good point! godbolt is amazing, I always forget to use it and use `grep`/Hopper :P

PR #56, opened 2019-09-27
add http client decompressor

artemredkin:
Add client HTTP decompression support

### Motivation:
Many HTTP servers can send compressed responses to clients and it would be a great feature for `AsyncHTTPClient` to support it. But since we want to minizime usage of unsafe APIs in SSWG projects, I propose to consolidate interfacing with zlib to `nio-extras` since it already supports `zlib` compression.

### Modifications:
Added `HTTPResponseDecompressor` and accompanying tests

### Result:
Users (primarily `AsyncHTTPClient`) can now provide automatic response decompression support.


Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,198 @@
...
+        guard rc == Z_OK else {
+            throw DecompressionError.initializationError(rc)
+        }
+    }
+}
+
+extension z_stream {
+    mutating func inflatePart(input: inout ByteBuffer, output: inout ByteBuffer) throws {
+        try input.readWithUnsafeMutableReadableBytes { (dataPtr: UnsafeMutableRawBufferPointer) -> Int in
+            let typedPtr = dataPtr.baseAddress!.assumingMemoryBound(to: UInt8.self)
```
Lukasa: I think these uses of `assumingMemoryBound` are not quite right: we should probably use `bindMemory` instead.

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,198 @@
...
+
+            try self.inflatePart(to: &output)
+
+            return typedDataPtr.count - Int(self.avail_in)
+        }
+    }
+
+    private mutating func inflatePart(to buffer: inout ByteBuffer) throws {
+        try buffer.writeWithUnsafeMutableBytes { outputPtr in
+            let typedOutputPtr = UnsafeMutableBufferPointer(start: outputPtr.baseAddress!.assumingMemoryBound(to: UInt8.self), count: outputPtr.count)
```
Lukasa: Same here.

Tests/LinuxMain.swift
```diff
@@ -1,43 +1,10 @@
...
-         testCase(JSONRPCFramingContentLengthHeaderEncoderTests.allTests),
-         testCase(LengthFieldBasedFrameDecoderTest.allTests),
-         testCase(LengthFieldPrependerTest.allTests),
-         testCase(LineBasedFrameDecoderTest.allTests),
-         testCase(QuiescingHelperTest.allTests),
-         testCase(RequestResponseHandlerTest.allTests),
-         testCase(WritePCAPHandlerTest.allTests),
-    ])
-#endif
+XCTMain(tests)
```
Lukasa: What happened here?

artemredkin: `swift test --generate-linuxmain`

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: like this?

Tests/LinuxMain.swift
Lukasa: Why aren't we using the test generation script in this directory?

artemredkin: habit, I guess :), re-generated
sidenote: don't we want to remove all those scripts since we can generate `LinuxMain` using SPM?

artemredkin: It still re-generated a lot of files...

Lukasa: > sidenote: don't we want to remove all those scripts since we can generate LinuxMain using SPM?

Nope, it's too easy to forget and doesn't work on Linux so it can't be enforced in CI. 

> It still re-generated a lot of files...

Delete the manifest file.

artemredkin: Done

artemredkin:

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,197 @@
...
+        case .head(let head):
+            let algorithm: CompressionAlgorithm?
+            let contentType = head.headers[canonicalForm: "Content-Encoding"].first?.lowercased()
+            if contentType == "gzip" {
+                algorithm = .gzip
+            } else if contentType == "deflate" {
+                algorithm = .deflate
+            } else {
+                algorithm = nil
+            }
```
tomerd: maybe encapsulate into extension along the lines of `CompressionAlgorithm(header: head.headers[canonicalForm: "Content-Encoding"].first?.lowercased())`

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,197 @@
...
+
+            let length = head.headers[canonicalForm: "Content-Length"].first.flatMap { Int($0) }
+
+            if let algorithm = algorithm, let length = length {
+                do {
+                    try self.initializeDecoder(encoding: algorithm, length: length)
+                } catch {
+                    context.fireErrorCaught(error)
```
tomerd: should code flow continue after this point?

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,197 @@
...
+                    } catch {
+                        context.fireErrorCaught(error)
+                        return
+                    }
+                }
+            default:
+                context.fireChannelRead(data)
+            }
+        case .end:
+            deflateEnd(&self.stream)
```
tomerd: self.

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,197 @@
...
+        self.stream.zfree = nil
+        self.stream.opaque = nil
+
+        let window: Int32
+        switch encoding {
+        case .gzip:
+            window = 15 + 16
+        default:
+            window = 15
+        }
```
tomerd:
```
var window: Int32 = 15
if case .gzip = encoding {
  window += 16
}
```
more readable?

artemredkin: How about moving it to a property in `CompressionAlgorithm`?

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: what do you mean? To `deflateEnd`? It's a function in `zlib.h`, not a method

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: good catch, fixed, thanks!

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: good idea, refactored, thanks!

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,203 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import CNIOExtrasZlib
+import NIO
+import NIOHTTP1
+
+/// Specifies how to limit decompression inflation.
+public enum DecompressionLimit {
```
weissi: please nest this type

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,203 @@
...
+            return false
+        case .size(let allowed):
+            return compressed > allowed
+        case .ratio(let ratio):
+            return decompressed > compressed * ratio
+        }
+    }
+}
+
+public enum DecompressionError: Error {
```
weissi: please nest this type

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,203 @@
...
+    case limit
+    case inflationError(Int32)
+    case initializationError(Int32)
+}
+
+public final class HTTPResponseDecompressor: ChannelDuplexHandler, RemovableChannelHandler {
```
weissi: the name of this class needs to start with `NIO` as per https://github.com/apple/swift-nio/blob/master/docs/public-api.md

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,203 @@
...
+            return compressed > allowed
+        case .ratio(let ratio):
+            return decompressed > compressed * ratio
+        }
+    }
+}
+
+public enum DecompressionError: Error {
+    case limit
+    case inflationError(Int32)
```
weissi: please use `Int` instead of `Int32`

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
weissi: Both `bindMemory` and `assumingMemoryBound` aren't optimal and almost equally (in)correct.

The reason they are equally correct or incorrect is:
- You may only `bind` memory if the memory is uninitialised or initialised to a type that is layout compatible with `UInt8`. The memory is definitely not uninitialised, so for `bind` to be correct you MUST KNOW that it's bound to something that is compatible with `UInt8`.
- `assumingMemoryBound` is only correct if you KNOW that the memory is bound to the type you supply.

Unless the user rebinds the memory, `ByteBuffer` is always bound to `UInt8` which makes both `bind` and `assumingMemoryBound` equally (in)correct: If the user hasn't rebound the memory it's correct; if the user has, then it's incorrect :).

Now, C has some extra special rule that you may interpret any pointer as a `(unsigned) char *` but Swift does not.

The right solution here is to shim the C function's interface in a way that it takes `(const) void *`s instead of `(const) (unsigned) char *`s as [we do here](https://github.com/apple/swift-nio/blob/master/Sources/CNIOHTTPParser/include/CNIOHTTPParser.h#L22-L27).

CC @atrick who I have chatted about this a number of times for him to double check.

Tests/LinuxMain.swift
weissi: It's a shame. Swift 5.1 can now (on Linux) also auto-discover the tests, but Swift 5.0 can not :(

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: it's not a function, it's a field in `z_stream_s` struct that we set, what would be correct way to assign it? create a function that accepts pointer to a struct and a `(const) void *`?

Lukasa: > You may only bind memory if the memory is uninitialised or initialised to a type that is layout compatible with UInt8. The memory is definitely not uninitialised, so for bind to be correct you MUST KNOW that it's bound to something that is compatible with UInt8.

I think we really need @atrick to write down the specific answer to this question: when a C function takes a `uint8_t *`, and you want to pass an `UnsafeRawBufferPointer` extracted from a `Data` or `ByteBuffer` to it, how do you do that? Is the answer _really_ “write a C shim for every such function”? If it is then frankly I think that needs to turn into a Swift bug report because the ergonomics of that are entirely horrible.

It was my understanding that binding the raw pointer retrieved from `withUnsafeBytes` is entirely acceptable so long as the memory accesses do not violate strict aliasing rules. If that’s not the case then I’m pretty concerned about the ergonomics.

weissi: @artemredkin this is extremely silly but the right thing to do here is:

```
#include <zlib.h>

static inline void *CNIOExtrasZLib_BytefPtr_to_voidPtr(Bytef *in) {
    return (void *)in;
}

static inline void *CNIOExtrasVoidPtr_to_BytefPtr(void *in) {
    return (Bytef *)in;
}
```

and dump that in [https://github.com/apple/swift-nio-extras/blob/master/Sources/CNIOExtrasZlib/include/CNIOZlib.h](CNIOZlib.h)

Oh, by the way, it's important that we rename `CNIOZlib.h` to `CNIOExtrasZlib.h`. For SwiftPM's xcode generation support it's _crucial_ that the module name and the umbrella header match _exactly_ (case matters).

artemredkin: done

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: good idea, fixed

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: good idea, fixed

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: renamed

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: done

artemredkin:
btw, do you know if CI is working? I haven't seen any successes or failures at all...

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
weissi: @Lukasa agreed, it’s a usability issue and feels pretty silly but creating such a shim was exactly what @atrick recommended to me. Coincidentally, that’s what we did for http_parser.

weissi:

weissi:
@artemredkin think you might not have been on the auto-trigger list yet

weissi:

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
atrick: Any time a Swift developer finds themselves needing to use either `bindMemory` or `assumingMemoryBound(to:)` to interoperate with C, I consider that a bug. `withMemoryRebound(to:)` is also very undesirable but sometimes necessary to workaround badly typed C APIs.

As far as calling C functions that take `char *`, this Swift feature would fix the most common cases: SR-10246: Support limited implicit pointer conversion when calling C functions.

That doesn't help you though. `zstream` is one of the most difficult C APIs to interoperate with safely because the `next_in` field is imported as a typed pointer but the client needs to assign it at arbitrary places to some pointer into raw bytes. I can think of three approaches:

1. Create a C shim over zlib that only exposes type safe pointers. e.g. `next_in` could be exposed as `void *`, or don't expose `next_in` at all, exposing a high-level API instead. I don't know how viable this is, but it solves the problem once an for all with no pointer type manipulation in Swift.

2. Guarantee that ByteBuffer will never be bound to a type that might be layout incompatible with `UInt8`. Reading and writing non-UInt8 values via the `UnsafeRawPointer` `loadBytes/storeBytes` API is fine though. Then you can safely call `bindMemory` to access ByteBuffer as a typed `UInt8` pointer whenever you want. Once you have a typed pointer, temporary rebinding to any type using `withMemoryRebound(to:)` is also fine, because it will be bound back to the `UInt8`-compatible type at the end of scope.

3. Sadly, this one isn't implemented yet, but...

[[SR-11087] UnsafeRawBufferPointer.withMemoryRebound(to:)
“Downcast” from raw to typed pointer limited to a closure scope](https://bugs.swift.org/browse/SR-11087)

That API looks like it would actually work in your case because `next_in` is only alive within some local scope. If you enclosed that local scope in `UnsafeRawBufferPointer.withMemoryRebound(to:)`, then it would be totally safe.

But, since #3 isn't available, I think you need to decide between #1 and #2.

Lukasa: @atrick Can I press you on the discussion in SR-10246? In that bug you say:

> In the above example, `bindMemory` is only safe because Data was responsible for allocating the memory itself, and it will never vend a typed pointer into that memory of some other type.

Is this true only for objects that are _locally_ allocated, i.e. where the programmer has complete knowledge of all possible binding points? If the program accepted a `Data` from elsewhere, I gather from this discussion that `bindMemory` would become unsafe?

Assuming that’s true, (2) isn’t workable, as `ByteBuffer`s will vend interior raw pointers to users, who can arbitrarily bind them to whatever type they want.

What I’m trying to understand here is whether memory binding in Swift is an implicitly global, persistent phenomenon. Once a pointer has been bound, is it an error to rebind it, even if the two types in question are primitive types (except via the combo of `assumingMemoryBound` followed by `withMemoryRebound`).

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,199 @@
...
+import NIOHTTP1
+
+public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableChannelHandler {
+    public typealias InboundIn = HTTPClientResponsePart
+    public typealias InboundOut = HTTPClientResponsePart
+    public typealias OutboundIn = HTTPClientRequestPart
+    public typealias OutboundOut = HTTPClientRequestPart
+
+    /// Specifies how to limit decompression inflation.
+    public enum DecompressionLimit {
```
weissi: Should we make this a `struct DecompressionLimit` with an inner private `enum` so that we can extend it to have other limits later without breaking public API? CC @Lukasa

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
atrick: *An addressable memory location is only bound to one type at a time.* That's really all there is to it, and that's sufficient to avoid undefined behavior due to aliasing pointers.

[The model only becomes slightly complicated when we introduce aggregate types--then we need the concept of memory being bound to an element within a larger type, but that's beside the point]

So, by binding memory, you are actually rebinding the memory's type. You can do that indefinitely as long as the memory is either uninitialized or initialized to any layout compatible type.

That's not really a normal, expected thing to do though. The normal way to rebind memory would be to call `withMemoryRebound(to:)`. But if you look at the implementation, all that's doing is binding memory at the scope start and binding it again at the end.

Ideally, no one ever needs to call `bindMemory` directly unless they are writing their own memory allocator in Swift that reuses chunks for memory for different types. But when `bindMemory` is needed to work around a C API like zlib, there needs to be some higher level convention for coordinating the memory's type.

The whole problem with `Foundation.Data` is that it can be initialized without copying the memory. This means it doesn't necessarilly know who owns its memory! So, if it covertly binds that memory to another type and vends out a typed pointer, there's no way for the user of `Data` to know that some typed pointer they are still hanging onto is no longer valid.

For a library that provides an untyped "byte buffer" to clients, it is perfectly reasonable for the library to state: I'm providing this `ByteBuffer` to you, the client, but I "own" the memory and therefore control it's bound type. The client can get at the underlying `UnsafeRawPointer`, and they can reinterpret the memory using `loadBytes/storeBytes`, but they should never (re)bind the memory's type. 

You might be concerned that users of the library will call `bindMemory` anyway, even though they definitely should not do that. That's still ok though as long as they don't persist that typed pointer beyond another call into the library. If the library also never persists typed pointers, instead calling `bindMemory` whenever it needs one, then it's safe.

For the convention to work, the rule needs to be: don't persist a typed pointer across a potential call to `bindMemory` to a different type.

A stronger convention is: only the "owner" of the memory can call `bindMemory`.

The ideal convention that we want to keep working toward is: don't call `bindMemory`.

Lukasa: > For the convention to work, the rule needs to be: don't persist a typed pointer across a potential call to `bindMemory` to a different type.

Ok. 👍 I had been operating under the understanding that if a pointer was extracted from `withUnsafeBytes`, bound to `UInt8`, and passed into a C function that computes and then returns, there was no particular risk to that temporary binding. That seems to be what you're saying here.

However, as you point out, this isn't ideal, and the better convention is to expose shims to turn the C APIs into ones that take `void *`, at least until such time as further pointer options are available to remove that need.

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: refactored, thanks

artemredkin:
@Lukasa @weissi any additional comments?

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
atrick: Yeah, you can call `bindMemory` within a library and there's no immediate problem. The reason we say it's incorrect in general is that the caller of the library may have a typed pointer that persists across the library call.

artemredkin:
@Lukasa @weissi any additional comments?

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,207 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
weissi: copyright doesn't contain 2019. If this is a new thing, just put '2019', if this contains old code it should be 2017-2019

artemredkin: fixed!

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -0,0 +1,207 @@
...
+                    } catch {
+                        context.fireErrorCaught(error)
+                        return
+                    }
+                }
+            default:
+                context.fireChannelRead(data)
+            }
+        case .end:
+            deflateEnd(&self.stream)
```
Lukasa: This should be `inflateEnd`. Presumably also this shouldn't be unconditionally called, but only called when we have actually initialised the decoder?

artemredkin: good catch, fixed, thanks!

PR #59, opened 2019-10-01
Gzip request decompress

calebkleveter:
Adds a `NIOHTTPRequestDecompressor` channel handler to decompress incoming HTTP requests.

### Motivation:

There will be times when a client wishes to send larger requests with gzipped bodies to save on network traffic. This PR adds a `NIOHTTPRequestDecompressor` which can be added to the server's channel pipeline so those requests are automatically inflated.

### Modifications:

- Added a `CNIOExtrasZlib_voidPtr_to_BytefPtr` C method.
- Added a `NIOHTTPRequestDecompressor` type.
- Added a `HTTPResponseDecompressorTest` test case.

### Result:

Now you don't have to manually check the `Content-Encoding` header and decompress the body on each incoming request.


weissi:
@calebkleveter #60  is actually the shared code extraction. Mind checking if you think that'd work for you?

weissi:
@calebkleveter this should now be able to use the functionality from #60

calebkleveter:
@Lukasa I've updated the `NIOHTTPRequestDecompressor` implementation to use `NIOHTTPDecompression`.

Sources/NIOHTTPCompression/HTTPRequestDecompressor.swift
```diff
@@ -0,0 +1,83 @@
...
+                    try self.decompressor.decompress(part: &part, buffer: &buffer, originalLength: compression.contentLength)
+
+                    context.fireChannelRead(self.wrapInboundOut(.body(buffer)))
+                } catch let error {
+                    context.fireErrorCaught(error)
+                    return
+                }
+            }
+        case .end(let headers):
+            if self.compression != nil {
```
Lukasa: From a defensive programming perspective, this case should probably `nil` the decoder.

Sources/NIOHTTPCompression/HTTPRequestDecompressor.swift
```diff
@@ -0,0 +1,83 @@
...
+                self.decompressor.deinitializeDecoder()
+            }
+
+            context.fireChannelRead(self.wrapInboundOut(.end(headers)))
```
Lukasa: This uses the opposite pattern to line 57, which uses the original `data`. We should be consistent: either change line 57 to be like this or change this line to be like line 57.

Tests/NIOExtrasTests/XCTestManifests.swift
```diff
@@ -0,0 +1,195 @@
+#if !canImport(ObjectiveC)
```
Lukasa: We have a script checked into this repository that generates the tests, at `scripts/generate_linux_tests.rb`. Can you discard these manifests and use that script instead?

calebkleveter:
@Lukasa Fixed. 👍

Lukasa:

PR #60, opened 2019-10-02
extract common classes for server request decompressor

artemredkin:
Extract common decompression functions and types for future HTTP Request Decompressor

### Motivation:
Changes introduced in #56 could be used as a baseline for #59, but in order for second PR to be merged we need to extract common functionality and common types first.

### Modifications:
Introduced a namespace for all common public types: `NIOHTTPDecompression`
Introduced a common decompression utility type: `Decompressor` that can be reused between response and request decompressors

### Result:
Request decompression PR should be much simpler now.

### DIFF
--- a/Sources/NIOHTTPCompression/HTTPDecompression.swift
+++ b/Sources/NIOHTTPCompression/HTTPDecompression.swift
@@ -75,45 +75,44 @@ public enum NIOHTTPDecompression {
             }
         }
     }
-}
 
-struct Decompressor {
-    private let limit: NIOHTTPDecompression.DecompressionLimit
-    private var stream = z_stream()
-    private var inflated = 0
+    struct Decompressor {
+        private let limit: NIOHTTPDecompression.DecompressionLimit
+        private var stream = z_stream()
+        private var inflated = 0
 
-    init(limit: NIOHTTPDecompression.DecompressionLimit) {
-        self.limit = limit
-    }
+        init(limit: NIOHTTPDecompression.DecompressionLimit) {
+            self.limit = limit
+        }
 
-    mutating func decompress(part: inout ByteBuffer, buffer: inout ByteBuffer, originalLength: Int) throws {
-        try self.stream.inflatePart(input: &part, output: &buffer)
-        self.inflated += buffer.readableBytes
+        mutating func decompress(part: inout ByteBuffer, buffer: inout ByteBuffer, originalLength: Int) throws {
+            self.inflated += try self.stream.inflatePart(input: &part, output: &buffer)
 
-        if self.limit.exceeded(compressed: originalLength, decompressed: self.inflated) {
-            throw NIOHTTPDecompression.DecompressionError.limit
+            if self.limit.exceeded(compressed: originalLength, decompressed: self.inflated) {
+                throw NIOHTTPDecompression.DecompressionError.limit
+            }
+        }
+
+        mutating func initializeDecoder(encoding: NIOHTTPDecompression.CompressionAlgorithm, length: Int) throws {
+            self.stream.zalloc = nil
+            self.stream.zfree = nil
+            self.stream.opaque = nil
+
+            let rc = CNIOExtrasZlib_inflateInit2(&self.stream, encoding.window)
+            guard rc == Z_OK else {
+                throw NIOHTTPDecompression.DecompressionError.initializationError(Int(rc))
+            }
+        }
+
+        mutating func deinitializeDecoder() {
+            inflateEnd(&self.stream)
         }
     }
-
-    mutating func initializeDecoder(encoding: NIOHTTPDecompression.CompressionAlgorithm, length: Int) throws {
-        self.stream.zalloc = nil
-        self.stream.zfree = nil
-        self.stream.opaque = nil
-
-        let rc = CNIOExtrasZlib_inflateInit2(&self.stream, encoding.window)
-        guard rc == Z_OK else {
-            throw NIOHTTPDecompression.DecompressionError.initializationError(Int(rc))
-        }
-    }
-
-    mutating func deinitializeDecoder() {
-        inflateEnd(&self.stream)
-    }
 }
-
 
 extension z_stream {
-    mutating func inflatePart(input: inout ByteBuffer, output: inout ByteBuffer) throws {
+    mutating func inflatePart(input: inout ByteBuffer, output: inout ByteBuffer) throws -> Int {
+        var written = 0
         try input.readWithUnsafeMutableReadableBytes { pointer in
             self.avail_in = UInt32(pointer.count)
             self.next_in = CNIOExtrasZlib_voidPtr_to_BytefPtr(pointer.baseAddress!)
@@ -125,14 +124,15 @@ extension z_stream {
                 self.next_out = nil
             }
 
-            try self.inflatePart(to: &output)
+            written += try self.inflatePart(to: &output)
 
             return pointer.count - Int(self.avail_in)
         }
+        return written
     }
 
-    private mutating func inflatePart(to buffer: inout ByteBuffer) throws {
-        try buffer.writeWithUnsafeMutableBytes { pointer in
+    private mutating func inflatePart(to buffer: inout ByteBuffer) throws -> Int {
+        return try buffer.writeWithUnsafeMutableBytes { pointer in
             self.avail_out = UInt32(pointer.count)
             self.next_out = CNIOExtrasZlib_voidPtr_to_BytefPtr(pointer.baseAddress!)
 
--- a/Sources/NIOHTTPCompression/HTTPDecompression.swift
+++ b/Sources/NIOHTTPCompression/HTTPDecompression.swift
@@ -18,13 +18,13 @@ import NIO
 public enum NIOHTTPDecompression {
     /// Specifies how to limit decompression inflation.
     public struct DecompressionLimit {
-        enum Limit {
+        private enum Limit {
             case none
             case size(Int)
             case ratio(Int)
         }
 
-        var limit: Limit
+        private var limit: Limit
 
         /// No limit will be set.
         public static let none = DecompressionLimit(limit: .none)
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -16,12 +16,7 @@ import CNIOExtrasZlib
 import NIO
 import NIOHTTP1
 
-public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableChannelHandler {
-    public typealias InboundIn = HTTPClientResponsePart
-    public typealias InboundOut = HTTPClientResponsePart
-    public typealias OutboundIn = HTTPClientRequestPart
-    public typealias OutboundOut = HTTPClientRequestPart
-
+public enum NIOHTTPDecompression {
```
weissi: could we move this to a separate file? Maybe `HTTPDecompression.swift`?

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -130,40 +124,55 @@ public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableC
...
+                self.decompressor.deinitializeDecoder()
             default:
                 break
             }
             context.fireChannelRead(data)
         }
     }
+}
+
+fileprivate struct Decompressor {
```
weissi: could we also move this to the shared file?

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: done!

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
artemredkin: done!

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,156 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import CNIOExtrasZlib
+import NIO
+
+public enum NIOHTTPDecompression {
+    /// Specifies how to limit decompression inflation.
+    public struct DecompressionLimit {
+        enum Limit {
```
Lukasa: Any reason not to make this private?

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,156 @@
...
+            case none
+            case size(Int)
+            case ratio(Int)
+        }
+
+        var limit: Limit
```
Lukasa: Or this?

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,156 @@
...
+struct Decompressor {
+    enum State {
+        case empty
+        case compressed(NIOHTTPDecompression.CompressionAlgorithm, Int)
+    }
+
+    let limit: NIOHTTPDecompression.DecompressionLimit
+    var state = State.empty
+    var stream = z_stream()
+    var inflated = 0
```
Lukasa: Does all this state need to be `internal`?

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,156 @@
...
+
+    public init(limit: NIOHTTPDecompression.DecompressionLimit) {
```
Lukasa: No point having a `public` initializer on an internal type.

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,156 @@
...
+        self.stream.opaque = nil
+
+        let rc = CNIOExtrasZlib_inflateInit2(&self.stream, encoding.window)
+        guard rc == Z_OK else {
+            throw NIOHTTPDecompression.DecompressionError.initializationError(Int(rc))
+        }
+    }
+
+    mutating func deinitializeDecoder() {
+        inflateEnd(&self.stream)
```
Lukasa: Why doesn't this check `.state`? Why doesn't it reset it either?

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,156 @@
...
+        try self.stream.inflatePart(input: &part, output: &buffer)
+        self.inflated += buffer.readableBytes
+
+        if self.limit.exceeded(compressed: originalLength, decompressed: self.inflated) {
+            throw NIOHTTPDecompression.DecompressionError.limit
+        }
+    }
+
+    mutating func initializeDecoder(encoding: NIOHTTPDecompression.CompressionAlgorithm, length: Int) throws {
+        self.state = .compressed(encoding, length)
```
Lukasa: I feel like this should check `.state` too

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,156 @@
...
+    var state = State.empty
+    var stream = z_stream()
+    var inflated = 0
+
+    public init(limit: NIOHTTPDecompression.DecompressionLimit) {
+        self.limit = limit
+    }
+
+    mutating func decompress(part: inout ByteBuffer, buffer: inout ByteBuffer, originalLength: Int) throws {
+        try self.stream.inflatePart(input: &part, output: &buffer)
```
Lukasa: This should probably check `.state`.

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: fixed

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: fixed

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: fixed

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: fixed

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: I made `state` be part of the handler, not decompressor, wdyt?

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: same here

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: same here

calebkleveter:
These abstractions looks like they'll work for https://github.com/apple/swift-nio-extras/pull/59. Good work! 👍

Sources/NIOHTTPCompression/HTTPDecompression.swift
Lukasa: That works for me.

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,147 @@
...
+    private var stream = z_stream()
+    private var inflated = 0
+
+    init(limit: NIOHTTPDecompression.DecompressionLimit) {
+        self.limit = limit
+    }
+
+    mutating func decompress(part: inout ByteBuffer, buffer: inout ByteBuffer, originalLength: Int) throws {
+        try self.stream.inflatePart(input: &part, output: &buffer)
+        self.inflated += buffer.readableBytes
```
Lukasa: This doesn't work the way you want if `buffer` already has readable bytes. With general code we need to keep track of what the original number of readable bytes was.

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,147 @@
...
+            case .deflate:
+                return 15
+            case .gzip:
+                return 15 + 16
+            }
+        }
+    }
+}
+
+struct Decompressor {
```
Lukasa: This struct should be moved into the namespace.

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -0,0 +1,147 @@
...
+    private let limit: NIOHTTPDecompression.DecompressionLimit
+    private var stream = z_stream()
+    private var inflated = 0
+
+    init(limit: NIOHTTPDecompression.DecompressionLimit) {
+        self.limit = limit
+    }
+
+    mutating func decompress(part: inout ByteBuffer, buffer: inout ByteBuffer, originalLength: Int) throws {
```
Lukasa: Some part of this function should call `reserveCapacity` on `buffer` to make sure we have some bytes to write into.

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: good idea, fixed, thanks!

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: good catch, fixed

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: what amount we should reserve? some multiplier of the source block size?

Lukasa: Yeah, 2x/3x is probably a good choice.

artemredkin: done!

PR #61, opened 2019-10-23
fix NIO 2.9.0 deprecations

artemredkin:
Fix usage of deprecated methods


Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -134,7 +134,7 @@ extension z_stream {
     }
 
     private mutating func inflatePart(to buffer: inout ByteBuffer) throws -> Int {
-        return try buffer.writeWithUnsafeMutableBytes { pointer in
+        return try buffer.writeWithUnsafeMutableBytes(minimumWritableBytes: 0) { pointer in
```
artemredkin: we reserve buffer capacity in both cases (see lines 89 in HTTPDecompression.swift and 313 in HTTPResponseCompressor.swft)

Lukasa: Would it be better to refactor this to use the new API instead of capacity reservation?

artemredkin: in decompressor we ensure capacity once (since we can approximately guess how much data will be written) and then write multiple times - do we really need to ensure it every time?
in compression we create a buffer with expected length from the start, so again, is there a reason to ensure capacity in this case?

Lukasa: If we already have enough capacity, ensuring it costs nothing. It is no more expensive than the current code.

artemredkin: done

Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -86,9 +86,7 @@ public enum NIOHTTPDecompression {
         }
 
         mutating func decompress(part: inout ByteBuffer, buffer: inout ByteBuffer, originalLength: Int) throws {
-            buffer.reserveCapacity(part.readableBytes * 2)
-
-            self.inflated += try self.stream.inflatePart(input: &part, output: &buffer)
+            self.inflated += try self.stream.inflatePart(input: &part, output: &buffer, minimumCapacity: part.readableBytes * 2)
```
Lukasa: `part` is passed `inout`, so we don't need the minimum capacity here: we can push it down into `inflatePart` .

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
```diff
@@ -314,7 +314,7 @@ private struct PartialHTTPResponse {
         var outputBuffer = allocator.buffer(capacity: bufferSize)
 
         // Now do the one-shot compression. All the data should have been consumed.
-        compressor.oneShotDeflate(from: &body, to: &outputBuffer, flag: flag)
+        compressor.oneShotDeflate(from: &body, to: &outputBuffer, minimumCapacity: bufferSize, flag: flag)
```
Lukasa: This can also be pushed down one level.

artemredkin: using `buffer.capacity`?

Sources/NIOHTTPCompression/HTTPDecompression.swift
artemredkin: good idea, fixed

PR #69, opened 2019-10-30
Add api checker script

tkrajacic:
Add the script that was bundled with swift-nio to this repo so we can run it on CI

### Motivation:

With this change it is possible to change the CI configuration to also run the API checker

### Modifications:

Add `check_no_api_breakages.sh` to scripts directory

### Result:

🥳

### Notes:

The script checks for a shasum `cefc4ee5bb7bcdb7cb5a7747efa178dab3c794d5` (possibly because it is easier than checking for a specific string). There should be a comment explaining which string produces this hash. I don't know.

### DIFF
--- a/scripts/check_no_api_breakages.sh
+++ b/scripts/check_no_api_breakages.sh
@@ -51,7 +51,7 @@ function usage() {
     echo >&2
     echo >&2 "Examples:"
     echo >&2
-    echo >&2 "Check between master and tag 2.1.1 of swift-nio-extras:"
+    echo >&2 "Check between master and tag 1.2.0 of swift-nio-extras:"
     echo >&2 "  $0 https://github.com/apple/swift-nio-extras master 1.2.0"
     echo >&2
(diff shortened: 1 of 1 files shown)


scripts/check_no_api_breakages.sh
```diff
@@ -0,0 +1,122 @@
...
+
+function usage() {
+    echo >&2 "Usage: $0 REPO-GITHUB-URL NEW-VERSION OLD-VERSIONS..."
+    echo >&2
+    echo >&2 "This script requires a Swift 5.1+ toolchain."
+    echo >&2
+    echo >&2 "Examples:"
+    echo >&2
+    echo >&2 "Check between master and tag 2.1.1 of swift-nio-extras:"
+    echo >&2 "  $0 https://github.com/apple/swift-nio-extras master 1.2.0"
```
Lukasa: Can you get the tag and comment to match here?

Lukasa:

PR #70, opened 2019-10-30
Clean up test command in docker config files

tkrajacic:
Only set the test command on the base config file and pass in the sanitizer argument for the 5.1 image (as 5.0 doesn't support the thread sanitizer)

### Motivation:

We want to run the thread sanitizer to find threading issues

### Modifications:

Remove the test command from the distro specific config files and only have it on the base config. Set the `SANITIZER_ARG` variable on the 5.1 specific config file.

### Result:

Thread sanitizer is run for the Swift 5.1 docker image tests

### DIFF
--- a/docker/docker-compose.1604.51.yaml
+++ b/docker/docker-compose.1604.51.yaml
@@ -10,5 +10,6 @@ services:
         swift_version: "5.1"
 
   test:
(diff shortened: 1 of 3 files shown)


Lukasa:

Lukasa:

PR #74, opened 2020-01-16
WritePCAPHandler: write outbound data on flush not write

weissi:
### Motivation:

`NIOWritePCAPHandler` right now writes the outbound data on `write` and not when actually written. In many scenarios, that's okay but it's not as close as it gets to what's sent on the wire.

### Modifications:

Write the data when the write promise succeeds.

### Result:

The emitted `.pcap` is closer to what's actually sent on the wire.

### DIFF
--- a/Sources/NIOExtras/WritePCAPHandler.swift
+++ b/Sources/NIOExtras/WritePCAPHandler.swift
@@ -361,33 +361,31 @@ extension NIOWritePCAPHandler: ChannelDuplexHandler {
     }
     
     public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
-        defer {
-            context.write(data, promise: promise)
-        }
-        guard self.closeState == .notClosing else {
-            return
-        }
-        
-        let data = self.unwrapInboundIn(data)
-        self.buffer.clear()
-        do {
-            var data = data
-            while var payloadToSend = self.takeSensiblySizedPayload(buffer: &data) {
-                try self.buffer.writePCAPRecord(.init(payloadLength: payloadToSend.readableBytes,
-                                                      src: self.localAddress(context: context),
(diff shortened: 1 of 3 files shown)


weissi:
CC @glbrntt , still WIP but getting there.

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -364,31 +365,44 @@ extension NIOWritePCAPHandler: ChannelDuplexHandler {
...
-        guard self.closeState == .notClosing else {
-            return
-        }
-        
+
         let data = self.unwrapInboundIn(data)
         self.buffer.clear()
         do {
             var data = data
+            var futureWrittenOutboundBytes = self.writtenOutboundBytes
```
weissi: this isn't right, needs to be persisted in the state too

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -364,31 +365,44 @@ extension NIOWritePCAPHandler: ChannelDuplexHandler {
...
-                                                                     sequenceNumber: self.writtenOutboundBytes,
+                                                                     sequenceNumber: futureWrittenOutboundBytes,
                                                                      srcPort: .init(self.localAddress(context: context).port!),
                                                                      dstPort: .init(self.remoteAddress(context: context).port!))))
-                self.writtenOutboundBytes += payloadToSend.readableBytes
+                futureWrittenOutboundBytes += payloadToSend.readableBytes
                 self.buffer.writeBuffer(&payloadToSend)
             }
-            self.writeBuffer(self.buffer)
+            self.pendingWrites.append(self.buffer)
```
weissi: probably, we want to buffer the `PCAPRecord` and not the raw bytes because we should set the timestamp in `flush` and not here.

PR #79, opened 2020-02-12
Decompression of HTTP responses that do not contain a Content-Length header

### LINKED ISSUE #78: NIOHTTPResponseDecompressor doesn't work without Content-Length header
### Expected behavior
Compressed responses that do not contain a `Content-Length` header should be decompressed.

### Actual behavior
If the `Content-Length` header is missing, the handler won't act on the body data and will directly pass compressed data to the next handlers

### Steps to reproduce
Configure a pipeline as usual with a NIOHTTPResponseDecompressor and give it a compressed response without a `Content-Length` header

Current version at the time of posting this issue is `1.4.0`

Independent of Swift/OS version

adtrevor:
### Motivation:

Currently, if the Content-Length header of an HTTP response is missing, the handler won't act on the body data and will directly pass compressed data to the next handlers

### Modifications:

This commit should fix this issue by dynamically updating body size as it reads it when no `Content-Length` header is specified.

### Result:
Fixes #78

### DIFF
--- a/Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
+++ b/Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
@@ -21,12 +21,26 @@ public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableC
     public typealias OutboundIn = HTTPClientRequestPart
     public typealias OutboundOut = HTTPClientRequestPart
 
-    private enum State {
-        case empty
-        case compressed(NIOHTTPDecompression.CompressionAlgorithm, Int)
+    private struct Compression {
+        var algorithm: NIOHTTPDecompression.CompressionAlgorithm
+        enum LengthSpecification {
+            case contentLength
+            case implicit
+        }
+        var lengthSpecification: LengthSpecification
+        var compressedLength: Int
+        
+        mutating func updateLength(with buffer: ByteBuffer) {
+            switch self.lengthSpecification {
+            case .contentLength:
+                break
+            case .implicit:
+                self.compressedLength += buffer.readableBytes
+            }
+        }
     }
 
-    private var state = State.empty
+    private var compression: Compression? = nil
     private var decompressor: NIOHTTPDecompression.Decompressor
 
     public init(limit: NIOHTTPDecompression.DecompressionLimit) {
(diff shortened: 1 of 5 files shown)


adtrevor:
Is there anything special to do if `Content-Length` *isn't* specified but the `Connection` is `Keep-Alive`? I assumed it is already managed by the HTTP response parsers handlers.

Tests/NIOHTTPCompressionTests/HTTPResponseDecompressorTest.swift
```diff
@@ -34,45 +34,57 @@ class HTTPResponseDecompressorTest: XCTestCase {
...
-        let body = ByteBuffer.of(bytes: [120, 156, 75, 76, 28, 5, 200, 0, 0, 248, 66, 103, 17])
-        do {
-            try channel.writeInbound(HTTPClientResponsePart.body(body))
-        } catch let error as NIOHTTPDecompression.DecompressionError {
-            switch error {
-            case .limit:
-                // ok
-                break
-            default:
+        let headersArray = [
```
adtrevor: Let me know if this is better to do this in a separate test, I preferred to use a loop to vary the headers as that's the only difference.

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -68,25 +86,27 @@ public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableC
...
-            case .compressed(_, let originalLength):
+            self.compression?.compressedLength += part.readableBytes
+            switch self.compression {
+            case .some(let compression):
                 while part.readableBytes > 0 {
+                    self.compression?.updateLength(with: part)
                     var buffer = context.channel.allocator.buffer(capacity: 16384)
                     do {
-                        try self.decompressor.decompress(part: &part, buffer: &buffer, originalLength: originalLength)
+                        try self.decompressor.decompress(part: &part, buffer: &buffer, compressedLength: compression.compressedLength)
```
Lukasa: This block of code inconsistently uses `compression` and `self.compression?`. It'd be better to be a bit consistent on this front.

fabianfett:
@adtrevor @Lukasa This branch has been baking for quite some time. I would love if it lands... To fix:
https://github.com/swift-server/async-http-client/issues/280

@adtrevor do you mind if I take it from here?

adtrevor:
> @adtrevor do you mind if I take it from here?

No problem! 🙂

PeterAdams-A:
This was incorporated into #93

PR #85, opened 2020-04-17
WritePCAPHandler: support logging more than 4GiB of data

weissi:
Motivation:

Previously, WritePCAPHandler would crash if more than 4GiB of data were
either received or sent through the same instance of the
WritePCAPHandler because of a UInt32 overflow representing the TCP
sequence/ACK numbers.

Modifications:

Make TCP sequence/ACK numbers wrap around correctly.

Result:

- now you can send/receive up to 16 EiB of data :P.
- fixes rdar://61887658


weissi:
@glbrntt hopefully fixed once #86  is in

PR #86, opened 2020-04-17
fix docker setup

weissi:
Motivation:

We were missing the 5.2 & 5.3 docker compose files and also the syntax
wasn't flexible enough to pull in the new nightlies.

Modifications:

- always specify the full image name
- add 5.2 & 5.3

Result:

More CI & newer Swifts.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -28,7 +28,7 @@ services:
 
   test:
     <<: *common
-    command: /bin/bash -xcl "swift test -Xswiftc -warnings-as-errors $${SANITIZER_ARG-}"
+    command: /bin/bash -xcl "cat /etc/lsb-release && swift -version && swift test -Xswiftc -warnings-as-errors $${SANITIZER_ARG-}"
 
   # util
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,7 @@
 ARG swift_version=5.0
 ARG ubuntu_version=bionic
-FROM swift:$swift_version-$ubuntu_version
+ARG base_image=swift:$swift_version-$ubuntu_version
+FROM $base_image
 # needed to do again after FROM due to docker limitation
 ARG swift_version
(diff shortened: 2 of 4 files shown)


docker/docker-compose.yaml
```diff
@@ -28,7 +28,7 @@ services:
 
   test:
     <<: *common
-    command: /bin/bash -xcl "swift test -Xswiftc -warnings-as-errors $${SANITIZER_ARG-}"
+    command: /bin/bash -xcl "cat cat /etc/lsb-release && swift -version && swift test -Xswiftc -warnings-as-errors $${SANITIZER_ARG-}"
```
glbrntt, line 31: 😸😸
```suggestion
    command: /bin/bash -xcl "cat /etc/lsb-release && swift -version && swift test -Xswiftc -warnings-as-errors $${SANITIZER_ARG-}"
```

weissi: omg, thanks @glbrntt , fixed

docker/docker-compose.yaml
```diff
@@ -28,7 +28,7 @@ services:
 
   test:
     <<: *common
-    command: /bin/bash -xcl "swift test -Xswiftc -warnings-as-errors $${SANITIZER_ARG-}"
+    command: /bin/bash -xcl "cat /etc/lsb-release && swift -version && swift test -Xswiftc -warnings-as-errors $${SANITIZER_ARG-}"
```
tomerd, line 31: debug commands intentional?

tomerd:
looks good, one question regarding debug commands

docker/docker-compose.yaml
weissi: yes, we need this everywhere, otherwise it's impossible to tell what version of Swift (especially for `master` snapshots) was actually run.

weissi: (same for the OS version, better to have that in there 100% clear)

PR #88, opened 2020-04-22
Added NIOHTTPRequestCompressor to compress requests

adam-fowler:
Added ChannelOutboundHandler for compressing request bodies.

### Motivation:

There is a response compressor, a response decompressor and a request decompressor. This completes the set

### Modifications:

Added new file HTTPRequestCompressor.swift that contains a ChannelOutboundHandler which compresses HTTP bodies as they are pass through the Channel. Moved common code between HTTPResponseCompressor and HTTPRequestCompressor into HTTPCompression.swift. Added tests for HTTPRequestCompressor.


Tests/NIOHTTPCompressionTests/HTTPRequestCompressorTest.swift
```diff
@@ -0,0 +1,376 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import CNIOExtrasZlib
+import NIO
+import NIOHTTP1
+@testable import NIOHTTPCompression
+
+struct PromiseArray {
```
weissi, line 21: sorry, that's a nit. Would you mind moving all this to the bottom? Just so that the uni test class is up top in the file. That makes it much easier to find the "run this test suite" diamond button in Xcode :)

Sources/NIOHTTPCompression/HTTPCompression.swift
```diff
@@ -0,0 +1,130 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import CNIOExtrasZlib
+import NIO
+
+public enum HTTPCompression {
```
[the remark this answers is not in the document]
adam-fowler: I originally had a NIO prefix but it ended up being the same name as the module. I'll think of another name. By the way the `HTTPResponseCompressor` doesnt have a NIO prefix. Maybe you want to fix that.

Sources/NIOHTTPCompression/HTTPCompression.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    public enum CompressionAlgorithm: String {
+        case gzip
```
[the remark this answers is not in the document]
adam-fowler: `CompressionAlgorithm` as it is used to define what compression method the handler uses so yes it has to stay public. Not totally sure what you mean about the `enum` within a `struct`. Have you got an example? Like this?
```
public struct CompressionAlgorithm {
    enum Algorithm: String {
        case gzip
        case deflate
    }
    let algorithm: Algorithm
    public static func gzip() { return CompressionAlgorithm(algorithm: .gzip) }
    public static func deflate() { return CompressionAlgorithm(algorithm: .deflate) 
}
```
Not sure what this gives us though

Sources/NIOHTTPCompression/HTTPCompression.swift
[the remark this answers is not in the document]
weissi: @adam-fowler `HTTPResponseCompressor` came in version 1.0.0, therefore doesn't need a prefix. It's only for additions that come _without_ a SemVer major.

Good point, we must now clash with the module name, maybe `NIOHTTPCompressionSettings` or something?

Sources/NIOHTTPCompression/HTTPCompression.swift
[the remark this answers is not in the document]
weissi: We should use a `public static var gzip = CompressionAlgorithm(algorithm: .gzip)`.

It gives us extensibility. If we ever want to add another case, with the `struct` we just can, with the `enum` we can't (without breaking SemVer major).

Sources/NIOHTTPCompression/HTTPCompression.swift
```diff
@@ -0,0 +1,130 @@
...
+        case deflate
+    }
+    
+    public enum CompressionError: Error {
+        case uncompressedWritesPending
```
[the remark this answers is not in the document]
adam-fowler: I must say converting this to a struct makes catching this error a little more awkward. Instead of
```
catch NIOHTTPCompressionSettings.CompressionError.uncompressedWritesPending
```
You have to replace it with 
```
catch let error as NIOHTTPCompressionSettings.CompressionError:
    if error == .uncompressedWritesPending {
         ...
```

adam-fowler:
Just realised I deleted a public symbol `HTTPResponseCompressor.CompressionError`, will reinstate

Sources/NIOHTTPCompression/HTTPCompression.swift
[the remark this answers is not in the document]
Lukasa: One note: `public static let`. `public static var` allows the setting to be overridden, with hilarious results.

Sources/NIOHTTPCompression/HTTPCompression.swift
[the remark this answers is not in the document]
Lukasa: It makes catching it awkward, but if we don't do it we can never add another error to this enum: we have to mint new enums every time. It's a bad scene.

Sources/NIOHTTPCompression/HTTPCompression.swift
```diff
@@ -0,0 +1,130 @@
...
+            precondition(rc == Z_OK, "Unexpected return from zlib init: \(rc)")
+        }
+
+        mutating func compress(inputBuffer: inout ByteBuffer, allocator: ByteBufferAllocator, finalise: Bool) -> ByteBuffer {
+            assert(isActive)
+            let flags = finalise ? Z_FINISH : Z_SYNC_FLUSH
+            guard inputBuffer.readableBytes > 0 || finalise == true else { return allocator.buffer(capacity: 0) }
+            // deflateBound() provides an upper limit on the number of bytes the input can
+            // compress to. We add 5 bytes to handle the fact that Z_SYNC_FLUSH will append
+            // an empty stored block that is 5 bytes long.
```
Lukasa, line 82: This comment does not match the surrounding code: nothing adds 5.

Sources/NIOHTTPCompression/HTTPCompression.swift
```diff
@@ -0,0 +1,130 @@
...
+                                                            count: outputPtr.count)
+            self.avail_out = UInt32(typedOutputPtr.count)
+            self.next_out = typedOutputPtr.baseAddress!
+            rc = deflate(&self, flag)
+            return typedOutputPtr.count - Int(self.avail_out)
+        }
+
+        return rc
+    }
+}
```
Lukasa, line 148: The fact that this extension moved between files makes it really hard to review if there are any meaningful differences between the two implementations. Can we restore the extension to its original location and remove the `private`?

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
```diff
@@ -302,7 +260,7 @@ private struct PartialHTTPResponse {
         body.reserveCapacity(initialBufferSize)
     }
 
-    mutating private func compressBody(compressor: inout z_stream, allocator: ByteBufferAllocator, flag: Int32) -> ByteBuffer? {
+    /*mutating private func compressBody(compressor: inout z_stream, allocator: ByteBufferAllocator, flag: Int32) -> ByteBuffer? {
```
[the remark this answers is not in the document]
adam-fowler: done

Sources/NIOHTTPCompression/HTTPCompression.swift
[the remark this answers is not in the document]
adam-fowler: done

Sources/NIOHTTPCompression/HTTPCompression.swift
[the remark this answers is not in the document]
adam-fowler: done

Sources/NIOHTTPCompression/HTTPCompression.swift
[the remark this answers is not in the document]
adam-fowler: done

Sources/NIOHTTPCompression/HTTPRequestCompressor.swift
```diff
@@ -0,0 +1,163 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import CNIOExtrasZlib
+import NIO
+import NIOHTTP1
+
+public final class HTTPRequestCompressor: ChannelOutboundHandler, RemovableChannelHandler {
```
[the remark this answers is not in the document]
adam-fowler: done

Tests/NIOHTTPCompressionTests/HTTPRequestCompressorTest.swift
```diff
@@ -0,0 +1,376 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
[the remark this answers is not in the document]
adam-fowler: done

Tests/NIOHTTPCompressionTests/HTTPRequestCompressorTest.swift
adam-fowler: done

Sources/NIOHTTPCompression/HTTPRequestCompressor.swift
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+    
+    public func flush(context: ChannelHandlerContext) {
+        switch state {
+        case .head(var head):
+            head.headers.remove(name: "Content-Length")
+            head.headers.replaceOrAdd(name: "Transfer-Encoding", value: "chunked")
+            context.write(wrapOutboundOut(.head(head)), promise: pendingWritePromise)
+            state = .partialBody(context.channel.allocator.buffer(capacity: 0))
```
Lukasa, line 146: I think there's a bug here.

Consider the scenario where the following three operations occur:

```
.write(.head)
.write(.body)
.flush
```

In this case we enter `func write` and transition `state` to `.head`. Then we enter `func write` again and transition state to `.body` _and add the `Content-Encoding` header_. Then we enter `flush`, remove content-length, add transfer-encoding, and write the data.

Consider now the alternative set of operations:

```
.write(.head)
.flush
.write(.body)
```

Here we'll start the same way: enter `func write` and transition `state` to `.head`. Then we'll enter `func flush`, which will send the head, never adding `Content-Encoding`.

Can we add a test case to cover this?

Sources/NIOHTTPCompression/HTTPCompression.swift
```diff
@@ -0,0 +1,148 @@
...
+
+            let rc = CNIOExtrasZlib_deflateInit2(&stream, Z_DEFAULT_COMPRESSION, Z_DEFLATED, windowBits, 8, Z_DEFAULT_STRATEGY)
+            precondition(rc == Z_OK, "Unexpected return from zlib init: \(rc)")
+        }
+
+        mutating func compress(inputBuffer: inout ByteBuffer, allocator: ByteBufferAllocator, finalise: Bool) -> ByteBuffer {
+            assert(isActive)
+            let flags = finalise ? Z_FINISH : Z_SYNC_FLUSH
+            guard inputBuffer.readableBytes > 0 || finalise == true else { return allocator.buffer(capacity: 0) }
+            // deflateBound() provides an upper limit on the number of bytes the input can
```
glbrntt, line 80: This is only true if `Z_FINISH` or `Z_NO_FLUSH` are used; from the zlib manual:

> Note that it is possible for the compressed size to be larger than the value returned by deflateBound() if flush options other than Z_FINISH or Z_NO_FLUSH are used.

We need to account for this because `oneShotDeflate` preconditions on the return code, which could be `Z_BUF_ERROR` if the output buffer isn't large enough.

Sources/NIOHTTPCompression/HTTPCompression.swift
```diff
@@ -0,0 +1,148 @@
...
+            // compress to. We add 5 bytes to handle the fact that Z_SYNC_FLUSH will append
+            // an empty stored block that is 5 bytes long.
```
glbrntt, line 82: The extra 5 bytes aren't added below. It's also it's not clear to me why adding 5 bytes would be okay here in light of the above comment; am I missing something?

Sources/NIOHTTPCompression/HTTPCompression.swift
adam-fowler: That was from the original `HTTPResponseCompressor` as it was common code I moved it out to a separate file. I guess by moving it I inherited it :-/. I'll go find out what happened here.

Sources/NIOHTTPCompression/HTTPRequestCompressor.swift
adam-fowler: fixed and added tests

Sources/NIOHTTPCompression/HTTPCompression.swift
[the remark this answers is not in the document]
weissi: @Lukasa thank you! I was thinking `public static var gzip: ... { ... }` but wrote it like a `let`. So bad that computed properties are always `var`. Teeny tiny little syntax change and suddenly you have the most hilarious API :P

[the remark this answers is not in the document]
adam-fowler: and I just copied it without thinking

Sources/NIOHTTPCompression/HTTPCompression.swift
weissi: You got me @adam-fowler , I noticed it was moved and skipped over it :P

Sources/NIOHTTPCompression/HTTPCompression.swift
adam-fowler: This was code moved across from HTTPResponseCompressor. I wanted to share code between the request and response compressors. I went to look for the original author of this and it was @weissi but his comment is copied to swift-nio-extras so the source of this is a little vague. I will look into zlib docs to see if I can get a better idea what this was for.

Sources/NIOHTTPCompression/HTTPCompression.swift
adam-fowler: @weissi skip comment below. Its your code, or at least you copied it into swift-nio-extras

Lukasa: I actually think it's _my_ code. At least, the comment reads like I wrote it.

I think the comment was true back when it was written, where we unconditionally used `Z_SYNC_FLUSH`. I believe the 5 bytes comment was the sense in which this can be larger that @glbrntt referred to: specifically, `Z_SYNC_FLUSH` may append an empty block, and that empty block is 5 bytes long.

What I don't know is how the code fell out of step with the comment. I'm not even 100% sure how true the comment is either.

adam-fowler: Found out why it's 5 bytes. From zlib docs
"If the parameter flush is set to Z_SYNC_FLUSH, all pending output is flushed to the output buffer and the output is aligned on a byte boundary, so that the decompressor can get all input data available so far. (In particular avail_in is zero after the call if enough output space has been provided before the call.) Flushing may degrade compression for some compression algorithms and so it should be used only when necessary. This completes the current deflate block and follows it with an empty stored block that is three bits plus filler bits to the next byte, followed by four bytes (00 00 ff ff). "

adam-fowler: I'll add in the comment and add the 5 bytes

adam-fowler: done

Sources/NIOHTTPCompression/HTTPCompression.swift
adam-fowler: See comment above

Sources/NIOHTTPCompression/HTTPCompression.swift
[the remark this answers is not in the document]
weissi: @adam-fowler yeah, the `switch` is unfortunate. In [this document](https://gist.github.com/weissi/d861bed9cad17871c3ad814911feeafb), I explored an alternative: 

```
} catch let error as NIOHTTPCompressionError where error == .uncompressedWritesPending {
```

not much nicer but...

Sources/NIOHTTPCompression/HTTPCompression.swift
```diff
@@ -0,0 +1,156 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import CNIOExtrasZlib
+import NIO
+
+public enum NIOHTTPCompressionSettings {
```
weissi, line 18: @adam-fowler / @Lukasa / @glbrntt so I proposed this `NIOHTTPCompressionSettings` as the namespace but then there are also errors in there so 'Settings' isn't that great.

Do we have better ideas that satisfy the following constraints:

- start with `NIO*`
- are not `NIOTHTTPCompression` (as that's the module name)

weissi: We could pull `NIOCompression` one and then call the nested ones just `public struct Algorithm` and `public struct Error`. So `NIOCompression.Algorithm`.

It's not perfect because obviously we may later want to compress stuff that's not HTTP but at least it reads nicely ;)

Lukasa: WFM.

Sources/NIOHTTPCompression/HTTPCompression.swift
glbrntt: I don't think adding 5 bytes revolves my comment above though: `deflateBound()` may underestimate the number of bytes required if anything other than `Z_FINISH` or `Z_NO_FLUSH` is used, it's not clear that adding the 5-bytes in case of `Z_SYNC_FLUSH` is sufficient. Given that this code already existed I'm happy if we solve this one separately to this PR.

Sources/NIOHTTPCompression/HTTPCompression.swift
adam-fowler: renamed

adam-fowler: Given the concern over possible later use of compression in NIO. Should the ByteBuffer compression be separated out from the the NIOHTTPCompression module and made for more general usage?

Sources/NIOHTTPCompression/HTTPCompression.swift
adam-fowler: @glbrntt This is the documentation telling you by how much it might over run in the case we use Z_SYNC_FLUSH. It discusses this elsewhere as well.

weissi:

Sources/NIOHTTPCompression/HTTPRequestCompressor.swift
```diff
@@ -0,0 +1,175 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import CNIOExtrasZlib
+import NIO
+import NIOHTTP1
+
+/// A HTTPResponseCompressor is a outbound channel handler that handles automatic streaming compression of
```
glbrntt, line 19:
```suggestion
/// NIOHTTPRequestCompressor is an outbound channel handler that handles automatic streaming compression of
```

Sources/NIOHTTPCompression/HTTPRequestCompressor.swift
```diff
@@ -0,0 +1,175 @@
...
+/// highly-compressible content, which will see the biggest benefits from streaming compression.
+public final class NIOHTTPRequestCompressor: ChannelOutboundHandler, RemovableChannelHandler {
+    public typealias OutboundIn = HTTPClientRequestPart
+    public typealias OutboundOut = HTTPClientRequestPart
+
+    /// Handler state
+    enum State {
+        /// handler hasn't started
+        case idle
+        /// handler has recived a head
```
glbrntt, line 37:
```suggestion
        /// handler has received a head
```

PR #93, opened 2020-07-01
Support decompression of HTTP responses that do not contain a Content…

### LINKED ISSUE #78: NIOHTTPResponseDecompressor doesn't work without Content-Length header

fabianfett:
This builds on top of #79.

- Rebased (Changed test structure to use `XCTAssertThrowsError`)
- Addressed @Lukasa's review comments

Original PR message from @adtrevor:

### Motivation:

Currently, if the Content-Length header of an HTTP response is missing, the handler won't act on the body data and will directly pass compressed data to the next handlers

### Modifications:

This commit should fix this issue by dynamically updating body size as it reads it when no Content-Length header is specified.

### Result:

Fixes #78

### DIFF
--- a/Sources/NIOHTTPCompression/HTTPDecompression.swift
+++ b/Sources/NIOHTTPCompression/HTTPDecompression.swift
@@ -46,7 +46,7 @@ public enum NIOHTTPDecompression {
         }
     }
 
-    public enum DecompressionError: Error {
+    public enum DecompressionError: Error, Equatable {
         case limit
         case inflationError(Int)
         case initializationError(Int)
@@ -86,15 +86,15 @@ public enum NIOHTTPDecompression {
             self.limit = limit
         }
 
-        mutating func decompress(part: inout ByteBuffer, buffer: inout ByteBuffer, originalLength: Int) throws {
+        mutating func decompress(part: inout ByteBuffer, buffer: inout ByteBuffer, compressedLength: Int) throws {
             self.inflated += try self.stream.inflatePart(input: &part, output: &buffer)
 
-            if self.limit.exceeded(compressed: originalLength, decompressed: self.inflated) {
+            if self.limit.exceeded(compressed: compressedLength, decompressed: self.inflated) {
                 throw NIOHTTPDecompression.DecompressionError.limit
             }
         }
 
-        mutating func initializeDecoder(encoding: NIOHTTPDecompression.CompressionAlgorithm, length: Int) throws {
+        mutating func initializeDecoder(encoding: NIOHTTPDecompression.CompressionAlgorithm) throws {
             self.stream.zalloc = nil
             self.stream.zfree = nil
             self.stream.opaque = nil
--- a/Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
+++ b/Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
@@ -21,12 +21,26 @@ public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableC
     public typealias OutboundIn = HTTPClientRequestPart
     public typealias OutboundOut = HTTPClientRequestPart
 
-    private enum State {
-        case empty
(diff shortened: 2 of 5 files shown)


Sources/NIOHTTPCompression/HTTPDecompression.swift
```diff
@@ -46,7 +46,7 @@ public enum NIOHTTPDecompression {
         }
     }
 
-    public enum DecompressionError: Error {
+    public enum DecompressionError: Error, Equatable {
```
fabianfett, line 49: 👀 @Lukasa This makes the tests a little easier to read. If this is an unwanted api change, I'm happy to revert it.

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -68,25 +86,27 @@ public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableC
 
             context.fireChannelRead(data)
         case .body(var part):
-            switch self.state {
-            case .compressed(_, let originalLength):
+            switch self.compression {
+            case .some(var compression):
+                compression.compressedLength += part.readableBytes
```
Lukasa, line 91: This can't be right: we set the `compressedLength` above differently depending on what mode we're in, but we are unconditionally adding to it now. That seems not right.

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -68,25 +86,27 @@ public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableC
...
                     context.fireChannelRead(self.wrapInboundOut(.body(buffer)))
                 }
-            default:
+            case .none:
                 context.fireChannelRead(data)
             }
         case .end:
-            switch self.state {
-            case .compressed:
+            switch self.compression {
```
Lukasa, line 108: Easier to rewrite this as `if self.compression != nil`.

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
Lukasa: Yeah, this looks like it is just leftover from a previous version of the patch. I think this indicates we need some extra tests to ensure that our compression limits still apply in the face of having no clear length specification here. Mind doing that?

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -68,25 +86,27 @@ public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableC
 
             context.fireChannelRead(data)
         case .body(var part):
-            switch self.state {
-            case .compressed(_, let originalLength):
+            switch self.compression {
```
Lukasa, line 89: This is now `if var compression`

Sources/NIOHTTPCompression/HTTPResponseDecompressor.swift
```diff
@@ -68,25 +86,27 @@ public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, RemovableC
...
+            case .some(var compression):
```
Lukasa, line 90: You aren't assigning the `compression` object back here, so these changes are all done to a copy which is thrown away. This is another reason to add the tests for the compression ratio policing: they'd have revealed this bug too.

Lukasa:

PR #94, opened 2020-07-20
Capture packets leading unto a user triggered event.

PeterAdams-A:
### Motivation:

Capturing all packets is expensive.  Recording to a ring buffer and 
then outputting on a triggering event allows this cost to be reduced.

### Modifications:

Add a new handler - NIOPCAPRingCaptureHandler.
This derives from the existing NIOWritePCAPHandler and generates PCAP recordings.
A ring buffer contained in this handler stores the captured packets until RecordPreviousPackets
is received as a user message at which point they are flushed to the sink.

### Result:

There is a new handler capable of outputting packet captured data only in the build up to 
a known event.


Lukasa:
I think we make `RingBuffer` public, we have to change its name for NIO compatibility reasons (and because we've greatly limited it, it's not a truly generic data structure, and it's also confusingly close to `CircularBuffer`), we give it a teeny tiny API and then yeah, we just push most of the rest into examples.

PeterAdams-A:
OK, I'll make that change - I actually toyed with that to start with and thought it felt slightly hard work to use but it's definitely more flexible.  I'll make the change quickly.

Lukasa:
We can provide a helper `static` method on `NIOWritePCAPHandler` that glues the ring buffer and handler together and gives them both back to you.

The thing that is most annoying is working out where you _store_ the ring buffer. In general though the right answer depends on the architecture of your application. If you're very pipeline-focused you can stuff it in a do-nothing handler you just summon out of the pipeline whenever you need it. If you've got a state object somewhere else, though, you probably have a better time putting it there making it easy to get to.

Lukasa:

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,81 @@
...
+    private var pcapFragments: CircularBuffer<ByteBuffer>
+    private var pcapCurrentBytes: size_t
+    private let maximumFragments: UInt
+    private let maximumBytes: size_t
+
+    /// Initialise the buffer, setting constraints.
+    /// Parameters:
+    ///    - maximumFragments: The maximum number of pcap fragments to store.
+    ///    - maximumBytes:  The maximum number of bytes to store - note, data written may exceed this by the header size.
+    public init(maximumFragments: UInt, maximumBytes: size_t) {
```
Lukasa, line 31: These should both be API as `Int`, and then converted to our preferred internal storage type internally.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,81 @@
...
+import NIO
+
+// MARK: NIOPCAPRingBuffer
+/// Storage for the most recent set of packets captured subject to constraints.
+/// Use `addFragment` as the sink to a `NIOWritePCAPHandler` and call `emitPCAP`
+/// when you wish to get the recorded data.
+public class NIOPCAPRingBuffer {
+    private var pcapFragments: CircularBuffer<ByteBuffer>
+    private var pcapCurrentBytes: size_t
+    private let maximumFragments: UInt
```
Lukasa, line 24: There's no benefit to storing this as `UInt`: we'll never use the top of the range anyway, and it forces type conversions for the comparison.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,81 @@
...
+        let popped = self.pcapFragments.popFirst()
+        if let popped = popped {
+            self.pcapCurrentBytes -= popped.readableBytes
+        }
+        return popped
+    }
+    
+    private func append(_ buffer: ByteBuffer) {
+        self.pcapFragments.append(buffer)
+        self.pcapCurrentBytes += buffer.readableBytes
```
Lukasa, line 49: Is it worth adding some defensive `assert`s in here to make sure we are maintaining our invariants?

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    /// Record a fragment into the buffer, making space if required.
+    /// Parameters:
+    /// - buffer: ByteBuffer containing a pcap fragment to store.
+    public func addFragment(_ buffer: ByteBuffer) {
+        // Make sure we don't go over on the number of fragments.
+        if self.pcapFragments.count >= self.maximumFragments {
+            self.popFirst()
+        }
+        assert(self.pcapFragments.count < self.maximumFragments)
```
Lukasa, line 60: Any reason for this not to be `precondition`?

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,81 @@
...
+        while self.pcapCurrentBytes > self.maximumBytes {
+            self.popFirst()
+        }
+    }
+
+    /// Emit the captured data to a byteBuffer - this drains the captured data.
+    /// Parameters:
+    /// - allocator: Allocator for creating byte buffers which are stored in the ring buffer.
+    public func emitPCAP(allocator: ByteBufferAllocator) -> ByteBuffer {
+        var buffer = allocator.buffer(capacity: self.pcapCurrentBytes)
```
Lukasa, line 75: Flattening the buffer here is not ideal: it'd be strictly better to offer this as a `CircularBuffer`, I suspect. Many places can tolerate the vector writes, and those that can't can always flatten this themselves.

We can just implement this by vending a copy of our circular buffer, calling `clear(keepingCapacity: true)`, and then setting the size to zero, which is much less computationally intensive.

Sources/NIOWritePartialPCAPDemo/main.swift
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/// Makes a series of http requests which get known responses.
+class SendSimpleSequenceRequestHandler: ChannelInboundHandler {
+    typealias InboundIn = HTTPClientResponsePart
+    typealias OutboundOut = HTTPClientRequestPart
+    
+    private let allDonePromise: EventLoopPromise<Void>
+
+    private var nextReqeustNumber = 0
```
Lukasa, line 55:
```suggestion
    private var nextRequestNumber = 0
```

Sources/NIOWritePartialPCAPDemo/main.swift
```diff
@@ -0,0 +1,136 @@
...
+                                                         .noContent, .resetContent, .preconditionFailed,
+                                                         .partialContent, .multiStatus, .alreadyReported ]
+
+    init(allDonePromise: EventLoopPromise<Void>) {
+        self.allDonePromise = allDonePromise
+    }
+    
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        if case .end = self.unwrapInboundIn(data) {
+            makeNextReqeustOrComplete(context: context)
```
Lukasa, line 66:
```suggestion
            self.makeNextRequestOrComplete(context: context)
```

Sources/NIOWritePartialPCAPDemo/main.swift
```diff
@@ -0,0 +1,136 @@
...
+    func errorCaught(context: ChannelHandlerContext, error: Error) {
+        self.allDonePromise.fail(error)
+        context.close(promise: nil)
+    }
+    
+    func channelActive(context: ChannelHandlerContext) {
+        makeNextReqeustOrComplete(context: context)
+    }
+
+    private func makeNextReqeustOrComplete(context: ChannelHandlerContext) {
```
Lukasa, line 79:
```suggestion
    private func makeNextRequestOrComplete(context: ChannelHandlerContext) {
```

Sources/NIOWritePartialPCAPDemo/main.swift
```diff
@@ -0,0 +1,136 @@
...
+        if self.nextReqeustNumber < self.requestsToMake.count {
```
Lukasa, line 80:
```suggestion
        if self.nextRequestNumber < self.requestsToMake.count {
```

Sources/NIOWritePartialPCAPDemo/main.swift
```diff
@@ -0,0 +1,136 @@
...
+            let headers = HTTPHeaders([("host", "httpbin.org"),
+                                       ("accept", "application/json")])
+            let currentStatus = self.requestsToMake[self.nextReqeustNumber].code
```
Lukasa, line 83:
```suggestion
            let currentStatus = self.requestsToMake[self.nextRequestNumber].code
```

Sources/NIOWritePartialPCAPDemo/main.swift
```diff
@@ -0,0 +1,136 @@
...
+            self.nextReqeustNumber += 1
```
Lukasa, line 84:
```suggestion
            self.nextRequestNumber += 1
```

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,81 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+// MARK: NIOPCAPRingBuffer
+/// Storage for the most recent set of packets captured subject to constraints.
+/// Use `addFragment` as the sink to a `NIOWritePCAPHandler` and call `emitPCAP`
+/// when you wish to get the recorded data.
+public class NIOPCAPRingBuffer {
```
Lukasa, line 21: This must note that the API is not thread-safe, and users are required to ensure that they only call it from the correct thread.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,81 @@
...
+    private var pcapFragments: CircularBuffer<ByteBuffer>
+    private var pcapCurrentBytes: size_t
+    private let maximumFragments: UInt
+    private let maximumBytes: size_t
+
+    /// Initialise the buffer, setting constraints.
+    /// Parameters:
+    ///    - maximumFragments: The maximum number of pcap fragments to store.
+    ///    - maximumBytes:  The maximum number of bytes to store - note, data written may exceed this by the header size.
```
glbrntt, line 30: Do we want to support both conditions at once? It feels like limiting by `maximumFragments` OR `maximumBytes` would be more useful.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,81 @@
...
+/// Use `addFragment` as the sink to a `NIOWritePCAPHandler` and call `emitPCAP`
+/// when you wish to get the recorded data.
+public class NIOPCAPRingBuffer {
+    private var pcapFragments: CircularBuffer<ByteBuffer>
+    private var pcapCurrentBytes: size_t
+    private let maximumFragments: UInt
+    private let maximumBytes: size_t
+
+    /// Initialise the buffer, setting constraints.
+    /// Parameters:
```
glbrntt, line 28: nit: you need `/// - Parameters:` for Xcode to render the documentation it correctly (here and elsewhere) 

fyi: you can use "⌘⌥/" on the function signature to generate the boilerplate for you

Tests/NIOExtrasTests/PCAPRingBufferTest.swift
```diff
@@ -0,0 +1,202 @@
...
+        let expectedData = 150 + 25 + 75 + 120
+        let ringBuffer = NIOPCAPRingBuffer(maximumFragments: 1000, maximumBytes: expectedData + 10)
+        for fragment in dataForTests() {
+            ringBuffer.addFragment(fragment)
+        }
+        let emitted = ringBuffer.emitPCAP(allocator: ByteBufferAllocator())
+        XCTAssertEqual(emitted.readableBytes, expectedData)
+    }
+    
+    func testExtremeByteLimit() {
```
glbrntt, line 62: Probably worth testing on the limit as well (i.e. `maximumBytes: 120`)

Tests/NIOExtrasTests/PCAPRingBufferTest.swift
```diff
@@ -0,0 +1,202 @@
...
+        func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+            if bytesUntilTrigger > 0 {
+                self.bytesUntilTrigger -= self.unwrapOutboundIn(data).readableBytes
+                if self.bytesUntilTrigger <= 0 {
+                    let ourPromise = context.eventLoop.makePromise(of: Void.self)
+                    context.write(data, promise: ourPromise)
+                    ourPromise.futureResult.flatMap {
+                        self.sink(self.pcapRingBuffer.emitPCAP(allocator: context.channel.allocator))
+                        return ourPromise.futureResult
+                    }.cascade(to: promise)
```
glbrntt, line 149: isn't this just the following?

```swift
context.write(data).map {
  self.sink(...)
}.cascade(to: promise)
```

Sources/NIOExtras/PCAPRingBuffer.swift
glbrntt: Yeah, I think that would make the API clearer..

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,90 @@
...
+        if let popped = popped {
+            self.pcapCurrentBytes -= popped.readableBytes
+        }
+        return popped
+    }
+    
+    private func append(_ buffer: ByteBuffer) {
+        self.pcapFragments.append(buffer)
+        self.pcapCurrentBytes += buffer.readableBytes
+        assert(self.pcapFragments.count <= self.maximumFragments)
```
glbrntt, line 53: A comment/documentation noting that we don't maintain the maximum size invariant here would be useful.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /// Record a fragment into the buffer, making space if required.
+    /// - Parameters:
+    /// - buffer: ByteBuffer containing a pcap fragment to store.
```
glbrntt, line 58: nit: the parameters need to be indented as well:

```swift
/// - Parameters:
///   - buffer: ...
```

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,90 @@
...
+    public func emitPCAP(_ consumer: (CircularBuffer<ByteBuffer>) -> Void) {
+        consumer(self.pcapFragments)
+        self.pcapFragments.removeAll(keepingCapacity: true)
+        self.pcapCurrentBytes = 0
+
+     /*   var buffer = allocator.buffer(capacity: self.pcapCurrentBytes)
+        while var next = self.popFirst() {
+            buffer.writeBuffer(&next)
+        }
+        return buffer*/
```
glbrntt, line 88: This'll need to go!

Sources/NIOWritePartialPCAPDemo/main.swift
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+    
+    func errorCaught(context: ChannelHandlerContext, error: Error) {
+        self.allDonePromise.fail(error)
+        context.close(promise: nil)
+    }
+    
+    func channelActive(context: ChannelHandlerContext) {
+        makeNextRequestOrComplete(context: context)
```
glbrntt, line 82:
```suggestion
        self.makeNextRequestOrComplete(context: context)
```

Tests/NIOExtrasTests/PCAPRingBufferTest.swift
```diff
@@ -0,0 +1,224 @@
...
+        XCTAssertNoThrow(try channel.pipeline.addHandler(
+                            NIOWritePCAPHandler(mode: .client,
+                                                fileSink: pcapRingBuffer.addFragment),
+                name: "capture").flatMap {
+                    return channel.pipeline.addHandler(
+                        TriggerOnCumulativeSizeHandler(triggerBytes: trigger,
+                                                       pcapRingBuffer: pcapRingBuffer,
+                                                       sink: testRecordedBytes),
+                        name: "trigger")
+                }.wait())
```
glbrntt, line 214: The formatting here is pretty crazy.

Splitting this into adding the handlers and then waiting on that future would help clear this up a bit.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,90 @@
...
+        if self.pcapFragments.count >= self.maximumFragments {
+            self.popFirst()
+        }
+        precondition(self.pcapFragments.count < self.maximumFragments)
+        
+        // Add the new fragment
+        self.append(buffer)
+
+        // Trim if we've exceeded byte limit - this could remove multiple, and indeed all fragments.
+        while self.pcapCurrentBytes > self.maximumBytes {
```
Davidde94, line 70: If I'm reading this correctly, then this can be slightly improved to run in O(1) time by having something like `self.popFirst(self.pcapCurrentBytes - self.maximumBytes)`

Lukasa: Sadly, that's wrong: there is no requirement that each fragment is only one byte. There's no `O(1)` implementation of this with the data structure we have in hand.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,90 @@
...
+            self.popFirst()
+        }
+        precondition(self.pcapCurrentBytes <= self.maximumBytes)
+    }
+
+    /// Emit the captured data to a consuming function; then clear the captured data.
+    /// - Parameters:
+    /// - consumer: Function which will take the stored fragments and output.
+    public func emitPCAP(_ consumer: (CircularBuffer<ByteBuffer>) -> Void) {
```
Lukasa, line 79: We could just `return` the `CircularBuffer`: might be easier.

Sources/NIOWritePartialPCAPDemo/main.swift
```diff
@@ -0,0 +1,142 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOExtras
+import NIOHTTP1
+
+
+
```
Lukasa, line 20: Let's delete at least one, and maybe two, of these whitespace lines.

Tests/NIOExtrasTests/PCAPRingBufferTest.swift
PeterAdams-A: I have split - still looks pretty crazy but better than it was.

Sources/NIOWritePartialPCAPDemo/main.swift
PeterAdams-A: ok.

Sources/NIOExtras/PCAPRingBuffer.swift
PeterAdams-A: Yes, my brain is still drawn towards reference semantics.  Will go for return.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,98 @@
...
+import NIO
+
+// MARK: NIOPCAPRingBuffer
+/// Storage for the most recent set of packets captured subject to constraints.
+/// Use `addFragment` as the sink to a `NIOWritePCAPHandler` and call `emitPCAP`
+/// when you wish to get the recorded data.
+/// - Warning:  This class is not thread safe so should only be called from one thread.
+public class NIOPCAPRingBuffer {
+    private var pcapFragments: CircularBuffer<ByteBuffer>
+    private var pcapCurrentBytes: size_t
```
glbrntt, line 24: Any reason we can't use `Int` here and below?

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,98 @@
...
+
+        // Trim if we've exceeded byte limit - this could remove multiple, and indeed all fragments.
+        while self.pcapCurrentBytes > self.maximumBytes {
+            self.popFirst()
+        }
+        precondition(self.pcapCurrentBytes <= self.maximumBytes)
+    }
+
+    /// Emit the captured data to a consuming function; then clear the captured data.
+    /// - Parameter consumer: Function which will take the stored fragments and output.
```
glbrntt, line 91: This no longer matches the implementation

PeterAdams-A: Blast - you're really good at spotting these.  Will fix.

Sources/NIOExtras/PCAPRingBuffer.swift
PeterAdams-A: Sure - much as I hate Int as a type it does match what we're getting from ByteBuffer.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -0,0 +1,98 @@
...
+    /// - Parameters:
+    ///    - maximumFragments: The maximum number of pcap fragments to store.
+    ///    - maximumBytes:  The maximum number of bytes to store - note, data written may exceed this by the header size.
+    public init(maximumFragments: Int, maximumBytes: Int) {
+        precondition(maximumFragments > 0)
+        precondition(maximumBytes > 0)
+        self.maximumFragments = maximumFragments
+        self.maximumBytes = maximumBytes
+        self.pcapCurrentBytes = 0
+        self.pcapFragments = CircularBuffer(initialCapacity: .init(maximumFragments))
```
Lukasa, line 38: Nit: this `.init` now does nothing.

PR #95, opened 2020-07-20
Don't install jazzy on xenial

PeterAdams-A:
Motivation:

Ubuntu 16 ruby is too old to support latest cocoapods gem.

Modifications:

Don't install jazzy when on xenial

Result:

Docker image will now build - you need to use bionic images to build documentation.


Lukasa:
I think this is going to affect all our CI (I know it affects some others already, I've seen it hit). So we do need a solution.

Updating Ruby is not going to work because we are already installing latest. We could instead flip over to using rvm, I suppose: that may be the better choice.

PR #98, opened 2020-07-27
Performance testing for NIO PCAP logging

PeterAdams-A:
Motivation:

It's useful to know the overhead we could be adding by including
the PCAP handler.

Modifications:

Add a new executable based on the NIO performance testing executable.

Result:

There is a new executable which runs a short test of sending and receiving
data through the HTTP/1 handler using multiple eventloops and showing
three options.
1) Vanilla
2) With in memory PCAP never written to disk
3) With a disk based PCAP.

### DIFF
--- a/Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
+++ b/Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
@@ -23,7 +23,7 @@ class HTTP1ThreadedPCapPerformanceTest: HTTP1ThreadedPerformanceTest {
         var fileSink: NIOWritePCAPHandler.SynchronizedFileSink!
 
         func setUp() throws {
-            let outputFile = NSTemporaryDirectory() + "/" + ProcessInfo().globallyUniqueString
+            let outputFile = NSTemporaryDirectory() + "/" + UUID().uuidString
             self.fileSink = try NIOWritePCAPHandler.SynchronizedFileSink.fileSinkWritingToFile(path: outputFile) { error in
                 print("ERROR: \(error)")
                 exit(1)
--- a/Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
+++ b/Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
@@ -16,8 +16,6 @@ import NIO
 import NIOExtras
 import Foundation
 
-
-
 class HTTP1ThreadedPCapPerformanceTest: HTTP1ThreadedPerformanceTest {
     private class SinkHolder {
         var fileSink: NIOWritePCAPHandler.SynchronizedFileSink!
--- a/Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
+++ b/Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
@@ -48,7 +48,7 @@ class HTTP1ThreadedPCapPerformanceTest: HTTP1ThreadedPerformanceTest {
                    extraInitialiser: { channel in return addPCap(channel: channel) })
     }
 
-    private var sinkHolder: SinkHolder
+    private let sinkHolder: SinkHolder
 
     override func run() throws -> Int {
         // Opening and closing the file included here as flushing data to disk is not known to complete until closed.
--- a/Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
+++ b/Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
@@ -1,0 +1,63 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2020 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOExtras
+import Foundation
+
+
+
+class HTTP1ThreadedPCapPerformanceTest: HTTP1ThreadedPerformanceTest {
+    private class SinkHolder {
+        var fileSink: NIOWritePCAPHandler.SynchronizedFileSink!
+
+        func setUp() throws {
+            let outputFile = NSTemporaryDirectory() + "/" + ProcessInfo().globallyUniqueString
+            self.fileSink = try NIOWritePCAPHandler.SynchronizedFileSink.fileSinkWritingToFile(path: outputFile) { error in
+                print("ERROR: \(error)")
+                exit(1)
+            }
+        }
+
+        func tearDown() {
+            try! self.fileSink.syncClose()
+        }
+    }
+
+    init() {
+        let sinkHolder = SinkHolder()
+        func addPCap(channel: Channel) -> EventLoopFuture<Void> {
+            let pcapHandler = NIOWritePCAPHandler(mode: .client,
+                                                  fileSink: sinkHolder.fileSink.write)
+            return channel.pipeline.addHandler(pcapHandler, position: .first)
+        }
+
+        self.sinkHolder = sinkHolder
+        super.init(numberOfRepeats: 50,
+                   numberOfClients: System.coreCount,
+                   requestsPerClient: 500,
+                   extraInitialiser: { channel in return addPCap(channel: channel) })
+    }
+
+    private var sinkHolder: SinkHolder
+
+    override func run() throws -> Int {
+        // Opening and closing the file included here as flushing data to disk is not known to complete until closed.
+        try sinkHolder.setUp()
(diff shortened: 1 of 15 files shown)


Sources/NIOExtrasPerformanceTester/PerformanceUtils.swift
```diff
@@ -0,0 +1,150 @@
...
+
+public func measureAndPrint(desc: String, fn: () throws -> Int) rethrows -> Void {
+    print("measuring\(warning): \(desc): ", terminator: "")
+    let measurements = try measure(fn)
+    print(measurements.reduce(into: "") { $0.append("\($1), ") })
+}
+
+// MARK: Utilities
+
+final class SimpleHTTPServer: ChannelInboundHandler {
```
Lukasa, line 44: I think I'd like these utilities in a separate file, just to indicate how separate they are from the rest of the measurement infrastructure.

Sources/NIOExtrasPerformanceTester/main.swift
```diff
@@ -0,0 +1,139 @@
...
+        let pcapHandler = NIOWritePCAPHandler(mode: .client,
+                                              fileSink: fileSink.write)
+        return channel.pipeline.addHandler(pcapHandler, position: .first)
+    }
+
+    return testHttp1Performance(numberOfRepeats: 50,
+                                numberOfClients: System.coreCount,
+                                requestsPerClient: 500,
+                                extraInitialiser: { channel in return addPCap(channel: channel) })
+}
```
Lukasa, line 139: I'd prefer to see these tests use the new `Benchmark` class-based model rather than this pattern. One test per file is a lot nicer.

Sources/NIOExtrasPerformanceTester/PerformanceUtils.swift
```diff
@@ -0,0 +1,150 @@
...
+    typealias InboundIn = HTTPServerRequestPart
+    typealias OutboundOut = HTTPServerResponsePart
+
+    private var files: [String] = Array()
```
Davidde94, line 48: Super nit: It's more "normal" to just use `[]`

Davidde94, line 48:
```suggestion
    private var files: [String] = []
```

Sources/NIOExtrasPerformanceTester/PerformanceUtils.swift
```diff
@@ -0,0 +1,150 @@
...
+    private var seenEnd: Bool = false
+    private var sentEnd: Bool = false
+    private var isOpen: Bool = true
```
Davidde94, line 51: Another super nit: Manually specifying the types like this isn't usually done

Sources/NIOExtrasPerformanceTester/PerformanceUtils.swift
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// MARK: Utilities
+
+final class SimpleHTTPServer: ChannelInboundHandler {
+    typealias InboundIn = HTTPServerRequestPart
+    typealias OutboundOut = HTTPServerResponsePart
+
+    private var files: [String] = Array()
+    private var seenEnd: Bool = false
```
Davidde94, line 49:
```suggestion
    private var seenEnd = false
```

Sources/NIOExtrasPerformanceTester/PerformanceUtils.swift
```diff
@@ -0,0 +1,150 @@
...
+    private var sentEnd: Bool = false
```
Davidde94, line 50:
```suggestion
    private var sentEnd = false
```

Sources/NIOExtrasPerformanceTester/PerformanceUtils.swift
Davidde94, line 51:
```suggestion
    private var isOpen = true
```

Sources/NIOExtrasPerformanceTester/PerformanceUtils.swift
```diff
@@ -0,0 +1,150 @@
...
+        head.headers.add(name: "Content-Length", value: "\(self.bodyLength)")
+        for i in 0..<self.numberOfAdditionalHeaders {
+            head.headers.add(name: "X-Random-Extra-Header", value: "\(i)")
+        }
+        self.cachedHead = head
+
+        var body: [UInt8] = []
+        body.reserveCapacity(self.bodyLength)
+        for i in 0..<self.bodyLength {
+            body.append(UInt8(i % Int(UInt8.max)))
```
Davidde94, line 69: Could maybe make this more readable with
```suggestion
            body.append(UInt8(i % 255)) // 255 == UInt8.max
```

Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
```diff
@@ -0,0 +1,63 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOExtras
+import Foundation
+
+
+
```
Lukasa, line 20: Maybe we don't need three blank lines here?

Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        self.sinkHolder = sinkHolder
+        super.init(numberOfRepeats: 50,
+                   numberOfClients: System.coreCount,
+                   requestsPerClient: 500,
+                   extraInitialiser: { channel in return addPCap(channel: channel) })
+    }
+
+    private var sinkHolder: SinkHolder
```
Lukasa, line 53:
```suggestion
    private let sinkHolder: SinkHolder
```

Sources/NIOExtrasPerformanceTester/HTTP1PerformanceTestFramework.swift
```diff
@@ -0,0 +1,204 @@
...
+            }
+            let allWrites = EventLoopFuture<Void>.andAllComplete(writeFutures, on: writeFutures.first!.eventLoop)
+            try! allWrites.wait()
+
+            let streamCompletedFutures = requestHandlers.map { rh in rh.completedFuture }
+            let requestsServed = EventLoopFuture<Int>.reduce(0, streamCompletedFutures, on: streamCompletedFutures.first!.eventLoop, +)
+            reqs.append(try! requestsServed.wait())
+        }
+        return reqs.reduce(0, +) / self.numberOfRepeats
+    }
```
Lukasa, line 203: Ok, this test is doing a _lot_, and very little of it is what we want to measure.

We should investigate rewriting this benchmark in terms of `EmbeddedChannel`. That'll remove all the network overhead, all the threading context switching mess, and many of the future allocations.

Is there any reason not to make that change?

PeterAdams-A: I'll try doing this as an additional rather than replacement set of tests as you rightly point out they have value but test very different things.

Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
PeterAdams-A: ok.

Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
PeterAdams-A: maybe - deleted.

PeterAdams-A:
@Lukasa take another look at this if you get a sec.

Sources/NIOExtrasPerformanceTester/PCAPPerformaceTest.swift
```diff
@@ -0,0 +1,64 @@
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1: This file should probably be called `PCAPPerformanceTest`

Sources/NIOExtrasPerformanceTester/PCAPPerformaceTest.swift
```diff
@@ -0,0 +1,64 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOExtras
+import Foundation
+
+class PCapPerformanceTest: Benchmark {
```
Lukasa, line 19:
```suggestion
class PCAPPerformanceTest: Benchmark {
```

Sources/NIOExtrasPerformanceTester/RollingPCAPPerformanceTest.swift
```diff
@@ -0,0 +1,52 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOExtras
+
+class RollingPCapPerformanceTest: Benchmark {
```
Lukasa, line 18:
```suggestion
class RollingPCAPPerformanceTest: Benchmark {
```

PR #103, opened 2020-08-04
Get API breakage script working on Linux for swift-5.2+

PeterAdams-A:
Motivation:

Current script always reports sdk errors on 5.2.

Modifications:

Only give -sdk where really necessary on 5.2 linux.

Result:

Script will work on 5.2 (although still reports silent errors)


PeterAdams-A:
SR-11143 is even more difficult to work around in 5.2. You have to set -sdk for -dump-sdk phase (which will report errors now as not found).  Not setting it on -diagnose-sdk avoid errors being reported into the final result.

PR #104, opened 2020-08-04
Optimise "startsWithSameUnicodeScalars", fix documentation

karwa:
We don't need to bother decoding unicode scalars out of the strings: if we're not doing any normalisation, we can can just compare the scalars in their UTF8-encoded form. While I'm at it, use `StaticString` since this is only used to test for literals such as "gzip".

### Motivation:

Small optimisation.

### Modifications:

See above. Also, the previous documentation was wrong, talking about generic collections of `Unicode.Scalar`s.

### Result:

Code which uses this utility function should be a bit more efficient.

### DIFF
--- a/Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
+++ b/Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
@@ -17,15 +17,15 @@ import NIO
 import NIOHTTP1
 
 extension StringProtocol {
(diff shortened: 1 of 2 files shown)


karwa:
@Lukasa Sure. That was already part of this patch, so I just dropped the commit which switched to `StaticString`.

Lukasa:

PR #111, opened 2021-01-22
Use `eventLoop.assertInEventLoop()` over `assert(eventLoop.inEventLoop)`

fabianfett:
Found four rare occurrences of `assert(eventLoop.inEventLoop)`. Fixed them.

### Motivation:

There is some reasoning why we should use `assertInEventLoop()` over `assert(eventLoop.inEventLoop)` but I can neither recall it nor find anything with the search engine I trust.

### Modifications:

- Replaced with `assert(eventLoop.inEventLoop)` with `eventLoop.assertInEventLoop() `

### Result:

SwiftNIOer code.

### DIFF
--- a/Sources/NIOExtras/QuiescingHelper.swift
+++ b/Sources/NIOExtras/QuiescingHelper.swift
@@ -48,7 +48,7 @@ private final class ChannelCollector {
     /// - parameters:
     ///   - channel: The `Channel` to add to the `ChannelCollector`.
     func channelAdded(_ channel: Channel) throws {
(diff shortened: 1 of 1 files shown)


glbrntt:
@fabianfett this is the explanation you're looking for: https://github.com/apple/swift-nio-transport-services/blob/698e0eed4740385c2d0f556a57fbebf468ae10c5/Sources/NIOTransportServices/NIOTSEventLoop.swift#L77-L94

Lukasa (approved):

fabianfett:
@glbrntt thanks!

PR #114, opened 2021-02-11
add support for a 24 bit (3 byte) length field

dnadoba:
### Motivation:

[The RSocket protocol uses a 24 bit length field](https://rsocket.io/docs/protocol#framing-format)

### Modifications:

- add two new methods readInteger and writeInteger on `ByteBuffer` that support reading and writing integers of any size
- add a new case (`.three`) to `ByteLength`

### Result:

`LengthFieldBasedFrameDecoder` & `LengthFieldPrepender` do now support a 24 bit length field

#### Outlook

If this pull request is merged, I can add `.five`, `.six` and `.seven` in a separate pull request if it is desired.
Maybe the new methods on `ByteBuffer` are of general interest and they could also be made public or be included in the main repository.

### DIFF
--- a/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
+++ b/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
@@ -40,7 +40,7 @@ extension ByteBuffer {
     mutating func read24UInt(
         endianness: Endianness = .big
     ) -> UInt32? {
-        guard let integer = get24UInt(at: readerIndex, endianness: endianness) else { return nil }
+        guard let integer = get24UInt(at: self.readerIndex, endianness: endianness) else { return nil }
         self.moveReaderIndex(forwardBy: 3)
         return integer
     }
--- a/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
+++ b/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
@@ -13,6 +13,56 @@
 //===----------------------------------------------------------------------===//
 
 import NIO
+import Foundation
+
+extension FixedWidthInteger {
+    @inlinable
+    init<D>(bytes: D) where D: DataProtocol {
+        var integer = Self()
+        withUnsafeMutableBytes(of: &integer) { (pointer) in
+            _ = bytes.copyBytes(to: pointer, count: Swift.min(bytes.count, MemoryLayout<Self>.size))
+        }
+        self = integer
+    }
+    @inlinable
+    func toEndianness(endianness: Endianness) -> Self {
+        switch endianness {
+        case .little:
+            return self.littleEndian
+        case .big:
+            return self.bigEndian
+        }
+    }
+}
+
+
+extension ByteBuffer {
+    /// Read `size` bytes off this `ByteBuffer`, move the reader index forward by `size` bytes and converts it into an `Integer`.
+    /// - Parameters:
+    ///   - size: The number of bytes to be read from this `ByteBuffer`.
+    ///   - endianness: The endianness of the integer in this `ByteBuffer` (defaults to big endian).
+    ///   - as: the desired `FixedWidthInteger` type (optional parameter)
+    /// - returns: An integer value deserialised from this `ByteBuffer` or `nil` if there aren't enough bytes readable.
+    /// - precondition: `size` must be less or equal to the size of `Integer`
+    @inlinable
+    mutating func readInteger<Integer>(
+        size: Int,
+        endianness: Endianness = .big,
+        type: Integer.Type = Integer.self
+    ) -> Integer? where Integer: FixedWidthInteger {
+        precondition(size <= MemoryLayout<Integer>.size, "requested byte count does not fit into requested integer type")
+        return readBytes(length: size).map { bytes -> Integer in
+            let integer = Integer(bytes: bytes).toEndianness(endianness: endianness)
+            switch endianness {
+            case .little:
+                return integer
+            case .big:
+                let missingLeadingZerosBytes = MemoryLayout<Integer>.size - size
+                return integer >> (missingLeadingZerosBytes * UInt8.bitWidth)
+            }
+        }
+    }
+}
 
 ///
 /// A decoder that splits the received `ByteBuffer` by the number of bytes specified in a fixed length header
@@ -35,11 +85,11 @@ import NIO
 public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
     ///
     /// An enumeration to describe the length of a piece of data in bytes.
-    /// It is contained to lengths that can be converted to integer types.
     ///
     public enum ByteLength {
         case one
         case two
+        case three
         case four
         case eight
         
@@ -49,6 +99,8 @@ public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
                 return 1
             case .two:
                 return 2
+            case .three:
+                return 3
             case .four:
                 return 4
             case .eight:
@@ -172,6 +224,8 @@ public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
             return buffer.readInteger(endianness: self.lengthFieldEndianness, as: UInt32.self).map { Int($0) }
         case .eight:
             return buffer.readInteger(endianness: self.lengthFieldEndianness, as: UInt64.self).map { Int($0) }
+        case .three:
+            return buffer.readInteger(size: 3, endianness: self.lengthFieldEndianness, type: UInt32.self).map { Int($0) }
         }
     }
 }
--- a/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
+++ b/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
@@ -71,22 +71,15 @@ public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
     public enum ByteLength {
         case one
         case two
-        case three
         case four
         case eight
         
-        var length: Int {
+        fileprivate var bitLength: LengthFieldBitLength {
             switch self {
-            case .one:
-                return 1
-            case .two:
-                return 2
-            case .three:
-                return 3
-            case .four:
-                return 4
-            case .eight:
-                return 8
+            case .one: return .oneByte
+            case .two: return .twoBytes
+            case .four: return .fourBytes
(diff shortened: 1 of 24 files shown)


Lukasa (commented):
Thanks, this is a great patch!

Rather than adding the full generality, I wonder if it might be better to use the existing pieces of safe infrastructure to manage this. That is, we can read and write the length field as a combination of a UInt16 and a UInt8 where the endianness defines the order of the reads. 

How does that idea sound?

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -35,11 +85,11 @@ import NIO
 public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
     ///
     /// An enumeration to describe the length of a piece of data in bytes.
-    /// It is contained to lengths that can be converted to integer types.
     ///
     public enum ByteLength {
         case one
         case two
+        case three
```
weissi, line 92: Unfortunately, this is public API breaking 😭

weissi (changes requested):
I really like your patch but at the moment it's public API breaking. We need to make some ugly  hack to allow 24bit length fields...

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
weissi: I think we should have a new initialiser that takes a new "ByteLength" type that is a) extensible b) contains 24 bits as well. Maybe we can call it `BitLength` and do

```
public struct BitLength {
    private struct Backing {
        case bits8
       case bits16
        case bits24
        ...
    }
    private let bitLength: Backing

    public static oneByte: BitLength = BitLength(bitLength: .bits8)
    public static twoBytes: BitLength = BitLength(bitLength: .bits16)
    public static threeBytes: BitLength = BitLength(bitLength: .bits24)

    public static eithtBits: BitLength = BitLength(bitLength: .bits8)
    public static sixteenBits: BitLength = BitLength(bitLength: .bits16)
    ...
}
```

then we can also deprecate `ByteLength`. @Lukasa wdyt?

dnadoba:
> Thanks, this is a great patch!
> 
> Rather than adding the full generality, I wonder if it might be better to use the existing pieces of safe infrastructure to manage this. That is, we can read and write the length field as a combination of a UInt16 and a UInt8 where the endianness defines the order of the reads.
> 
> How does that idea sound?

Changed :)
I actually like the generic version more but reading/writing UInt16/UInt8 will probably be faster.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
dnadoba: Oh... I totally forgot about that.
@weissi, your proposed solution sounds good to me 👍

Sources/NIOExtras/LengthFieldBitLength.swift
```diff
@@ -0,0 +1,69 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
```
weissi: sorry for being annoying here but given that this is a new file, it should be just `(c) 2021`, unless I missed something and you actually took some code from elsewhere that had `2017 - ...`

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -13,6 +13,38 @@
 //===----------------------------------------------------------------------===//
 
 import NIO
+import Foundation
+
+extension ByteBuffer {
+    @inlinable
+    mutating func get24UInt(
+        endianness: Endianness = .big
```
weissi, line 21: `get*` functions typically take an index, so I'd suggest you add `index: Int` here and use that instead of the `readerIndex` below to be more in line with the other `get*` functions.

Sources/NIOExtras/LengthFieldPrepender.swift
```diff
@@ -13,6 +13,27 @@
 //===----------------------------------------------------------------------===//
 
 import NIO
+import Foundation
```
weissi, line 16: NIO packages don't usually depend on `Foundation` so they can be used where that's not appropriate. Any particular reason to `import Foundation` here?

Sources/NIOExtras/LengthFieldPrepender.swift
```diff
@@ -13,6 +13,27 @@
...
+
+extension ByteBuffer {
+    @discardableResult
+    @inlinable
+    mutating func write24UInt(
+        _ integer: UInt32,
+        endianness: Endianness = .big
+    ) -> Int {
+        assert(integer & 0xFF_FF_FF == integer, "integer value does not fit into 24 bit integer")
```
weissi, line 25: I'd make this a `precondition` so it's also checked in release mode.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -82,13 +109,23 @@ public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
     ///    - lengthFieldLength: The length of the field specifying the remaining length of the frame.
     ///    - lengthFieldEndianness: The endianness of the field specifying the remaining length of the frame.
     ///
-    public init(lengthFieldLength: ByteLength, lengthFieldEndianness: Endianness = .big) {
+    public convenience init(lengthFieldLength: ByteLength, lengthFieldEndianness: Endianness = .big) {
```
weissi, line 112: @Lukasa do we want to deprecate this one or just leave as is? Doesn't really cause harm

Sources/NIOExtras/LengthFieldBitLength.swift
```diff
@@ -0,0 +1,69 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// An struct to describe the length of a piece of data in bits
+public struct LengthFieldBitLength {
```
weissi, line 18: very annoyingly, we need to prefix this with `NIO*` or move it into one of the pre-existing types.

Sources/NIOExtras/LengthFieldBitLength.swift
```diff
@@ -0,0 +1,69 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
weissi, line 15: Foundation import

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -13,6 +13,38 @@
 //===----------------------------------------------------------------------===//
 
 import NIO
+import Foundation
```
weissi, line 16: Foundation import

weissi (changes requested):
Thanks! This looks really great, almost ready to go, just some cosmetics really.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
Lukasa: I'm fine with leaving it as-is.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
dnadoba: Good catch! All Foundation imports are leftovers from the initial generic implementation which used `Data` & `DataProtocol`.

Sources/NIOExtras/LengthFieldBitLength.swift
dnadoba: Changed, but why? `LengthFieldBitLength` is not an enum, but a struct.
How does this effect API stability?

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
dnadoba: Changed :)

Sources/NIOExtras/LengthFieldPrepender.swift
dnadoba: Done

Sources/NIOExtras/LengthFieldBitLength.swift
weissi: @dnadoba We [guarantee](https://github.com/apple/swift-nio/blob/main/docs/public-api.md#1-no-global-namespace-additions-that-arent-prefixed) that we won't add _anything_ that doesn't start with `NIO`/`CNIO`/... The reason for that is if a user writes this program:

```
import NIO
import SomethingElse

let l = LengthFieldBitField()
```

and let's assume that `SomethingElse` also provides a `LengthFieldBitField`, then the code would compile _until_ your PR gets merged and then suddenly `LengthFieldBitLength` becomes ambiguous. You, you can change it to `SomethingElse.LengthFieldBitLength` but that requires a code change and the compilation failed.

Therefore we guarantee to only add `NIO*` stuff and then this can't happen assuming that nobody else produces `NIO*` names.

weissi: What would obviously be awesome is if SwiftPM & Swift would only actually export the `NIO` API given the restrictions in `Package.swift`. If that were the case, then if you wrote `from: "2.0.0"` you would only get the NIO API introduced in 2.0.0 and not what came in 2.1.0 and later even if (say) you compiled against NIO 2.25.0. That would also solve the issue that people just put `from: "2.0.0"` but then end up using much newer NIO features.

But alas this feature isn't actually implemented so we have to use the `NIO*` prefixing hack and hope for the best :).

dnadoba: Thank you for your detailed explanation :)

weissi (approved):
Thank you! That looks great to me.

Lukasa (approved):

Lukasa:
Nice one, thanks!

PR #115, opened 2021-02-11
fix crash in LengthFieldBasedFrameDecoder for malicious length values

dnadoba:
### Motivation:

LengthFieldBasedFrameDecoder will cause a fatal error if the length value does not fit into an `Int`.
This can happen if `lengthFieldLength` is set to `.eight` and we are on a 64 bit platform or if `lengthFieldLength` is set to `.four` and we are on a 32-bit platform.
If we then receive a length field value which is greater than `Int.max` the conversion from `UInt` to `Int` will cause a fatal error.
This could be abused to crash a server by only sending 4 or 8 bytes.

### Modifications:

safely convert UInt64 & UInt32 to Int and throw an error if they can't be represented as an Int

### Result:

- LengthFieldBasedFrameDecoder with lengthFieldLength set to `.eight` can no longer crash the server on a 64-bit platform
- LengthFieldBasedFrameDecoder with lengthFieldLength set to `.four` can no longer crash the server on a 32-bit platform

### DIFF
--- a/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
+++ b/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
@@ -88,11 +88,6 @@ public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
     ///    - lengthFieldEndianness: The endianness of the field specifying the remaining length of the frame.
     ///
     public init(lengthFieldLength: ByteLength, lengthFieldEndianness: Endianness = .big) {
-
-        // The value contained in the length field must be able to be represented by an integer type on the platform.
-        // ie. .eight == 64bit which would not fit into the Int type on a 32bit platform.
-        precondition(lengthFieldLength.length <= Int.bitWidth/8)
-            
         self.lengthFieldLength = lengthFieldLength
         self.lengthFieldEndianness = lengthFieldEndianness
     }
--- a/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
+++ b/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
@@ -17,6 +17,8 @@ import NIO
 public enum LengthFieldBasedFrameDecoderError: Error {
     /// This error can be thrown by `LengthFieldBasedFrameDecoder` if the length field value is larger than `Int.max`
     case lengthFieldValueTooLarge
+    /// This error can be thrown by `LengthFieldBasedFrameDecoder` if the length field value is larger than `LengthFieldBasedFrameDecoder.maxSupportedLengthFieldSize`
+    case lengthFieldValueLargerThanMaxSupportedSize
 }
 
 ///
@@ -38,6 +40,8 @@ public enum LengthFieldBasedFrameDecoderError: Error {
 /// 'A' and 'E' will be the headers and will not be passed forward.
 ///
 public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
+    /// Maximum supported length field size in bytes of `LengthFieldBasedFrameDecoder` and is currently `Int32.max`
+    static let maxSupportedLengthFieldSize: Int = Int(Int32.max)
     ///
     /// An enumeration to describe the length of a piece of data in bytes.
     /// It is contained to lengths that can be converted to integer types.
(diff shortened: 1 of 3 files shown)


Lukasa (approved):
This looks like a great patch, thanks!

Lukasa:

Lukasa:
Unfortunately `XCTSkipIf` doesn't exist in the older Swift versions we support. I recommend just using an early exit instead.

Tests/NIOExtrasTests/LengthFieldBasedFrameDecoderTest.swift
```diff
@@ -455,4 +455,37 @@ class LengthFieldBasedFrameDecoderTest: XCTestCase {
...
+        if UInt32.bitWidth == UInt.bitWidth {
+            // LengthFieldBasedFrameDecoder.ByteLength.four should only throw if we are on a 32 bit platform
+            XCTAssertThrowsError(try self.channel.writeInbound(buffer))
+        } else {
+            XCTAssertFalse(try self.channel.writeInbound(buffer).isFull)
+        }
+    }
+    
+    func testMaliciousLengthOn64BitPlatform() {
+        // LengthFieldBasedFrameDecoder.ByteLength.eight is only supported on 64 bit systems
```
weissi, line 478: oh, why is that? It should be supported everywhere. If not, that's a bug we should fix.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -161,17 +166,27 @@ public final class LengthFieldBasedFrameDecoder: ByteToMessageDecoder {
     /// - parameters:
     ///    - buffer: The buffer containing the integer frame length.
     ///
-    private func readFrameLength(for buffer: inout ByteBuffer) -> Int? {
+    private func readFrameLength(for buffer: inout ByteBuffer) throws -> Int? {
 
         switch self.lengthFieldLength {
         case .one:
             return buffer.readInteger(endianness: self.lengthFieldEndianness, as: UInt8.self).map { Int($0) }
```
weissi, line 173: oh, that's not your fault but there are quite a few bugs here. We can't make an `Int` out of all of these values without potentially crashing. We need to store the length as `UInt64` on all platforms or establish some maximum. I'd be happy if we say the max supported is `UInt32.max` = 4 GB on all platforms.

weissi (changes requested):
Thank you so much for looking into this. There are quite a few bugs in the old code that you uncovered. I think your patch is mostly good but we should instead just store the length as a `UInt64` on all platforms internally.

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
Lukasa: We can't store the length in a `UInt64` because we can't pass that to `readSlice`. Having the Int fail to fit into UInt64 is fine: any slice that actually flips the top bit of a UInt64 is obviously ridiculous.

weissi: @Lukasa I see! Should we just say the maximum supported is 2GB and define that as the universal maximum on all platforms? Then we can continue to use `Int` and just need to check (as done in this PR). But we should still support 64 length fields on 32 bit.

Lukasa: SGTM.

dnadoba: I have removed the precondition check in the initialiser. 64 bit length fields are now supported on 32-bit platforms.
Should we really add an additional check and throw an error if a 64 bit length field exceeds 2GB on a 64-bit platform?
`ByteToMessageHandler` in apple/swift-nio already has the functionality to limit the maximum supported size.

weissi: @dnadoba I believe we should add this check for a couple of reasons:
- 32 and 64 bit APIs match and both support 2GB and no more
- the max we support in ByteBuffer is 4GB (BBs can't grow more than 4GB)

weissi: I should expand on the second point: Let's say the user were to allow up to 5 GB here (on 64bit platforms), then ByteBuffer's `write*` would crash on the `write*` that crosses 4 GB anyway. A user will not know this information so I think we should catch this here.

dnadoba: Got it, thanks for clarifying!

Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift
```diff
@@ -17,6 +17,8 @@ import NIO
 public enum LengthFieldBasedFrameDecoderError: Error {
     /// This error can be thrown by `LengthFieldBasedFrameDecoder` if the length field value is larger than `Int.max`
     case lengthFieldValueTooLarge
+    /// This error can be thrown by `LengthFieldBasedFrameDecoder` if the length field value is larger than `LengthFieldBasedFrameDecoder.maxSupportedLengthFieldSize`
+    case lengthFieldValueLargerThanMaxSupportedSize
```
weissi, line 21: this is also public API breaking 😭. You have two options:

1. just use `lengthFieldValueTooLarge` (which is what I would do)
2. create a new type inside `NIOExtraError.swift` called `LengthFieldValueLargerThanMaximumSupported` or similar

dnadoba: The enum is completely new in this pull request. The case was just added later. But if you like, I can still remove the separate enum case :)
BTW: Is there a recommended way of defining error's? If I remember correctly, only libraries which define ABI stability can add new enum cases to enum's which are not marked as `@forzen` but there was proposal to fix that kind of issue in non ABI stable libraries but wasn't accepted.

weissi: @dnadoba ah, that's great. In this case you just need to prefix the `enum` with `NIO` ie `NIOLengthFieldBased...` so we don't violate our [public API guarantees](https://github.com/apple/swift-nio/blob/main/docs/public-api.md#1-no-global-namespace-additions-that-arent-prefixed).

You are correct, that SwiftPM libraries don't have the option between "open" and `@frozen` enums sadly. All we've got are `@frozen` ones (without specifying that).

I'm happy with adding this `enum` however we could also move both into `NIOExtrasError`. Let's ask @Lukasa what he thinkgs

weissi:
@swift-nio-bot add to allowlist

Tests/NIOExtrasTests/LengthFieldBasedFrameDecoderTest.swift
```diff
@@ -455,4 +455,31 @@ class LengthFieldBasedFrameDecoderTest: XCTestCase {
...
+                                                                   lengthFieldEndianness: .little))
+        XCTAssertNoThrow(try self.channel.pipeline.addHandler(self.decoderUnderTest).wait())
+
+        let dataLength = UInt64(Int64.max) + 1
+        
+        var buffer = self.channel.allocator.buffer(capacity: 8) // 8 byte header
+        buffer.writeInteger(dataLength, endianness: .little, as: UInt64.self)
+        
+        XCTAssertThrowsError(try self.channel.writeInbound(buffer))
+    }
```
weissi, line 484: if you don't mind, we could add maybe two more edge-cases here which make sure we can correctly parse the maximum allowable value (`Int32.max`) and we fail with the minimum not-allowable (`Int32.max + 1`). For the second point, we can probably modify the tests above.

dnadoba: Sure :) added

weissi (approved):
thank you! Looks good to me!

PR #123, opened 2021-05-04
use 5.4 release instead of nightly

weissi:


Lukasa:
@tomerd can we add a 5.4 builder for this? Seems like it got overlooked.

Lukasa (approved):

yim-lee:
> can we add a 5.4 builder for this?

I will look into this

PR #130, opened 2021-06-08
stop publishing the executable products

weissi:


PeterAdams-A (approved):
Seems reasonable - I assume you've started using this as a library source.

PR #131, opened 2021-06-10
Fix SOCKS client sending greeting and port/address endianness

Davidde94:
The SOCKS client doesn't work in its current state. The client channel handler may be added upon channel creation, or later. To accommodate for this we attempt to begin the handshake in both `channelActive` and `handlerAdded`. `handlerAdded` is always called first, but if the handler is added before the channel is active then the client greeting will fail, and the state machine will enter an irrecoverable error state.

~This wasn't caught by the tests as the tests use an `EmbeddedChannel`, where the channel is always active.~

Modifications:
Guard the channel active state.

### DIFF
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
@@ -99,7 +99,7 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
 extension SOCKSClientHandler {
     
     private func beginHandshake(context: ChannelHandlerContext) {
-        guard self.state.shouldBeginHandshake else {
+        guard context.channel.isActive, self.state.shouldBeginHandshake else {
             return
         }
         do {
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
@@ -99,7 +99,7 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
 extension SOCKSClientHandler {
     
     private func beginHandshake(context: ChannelHandlerContext) {
-        guard context.channel.isActive,  self.state.shouldBeginHandshake else {
+        guard context.channel.isActive, self.state.shouldBeginHandshake else {
             return
         }
         do {
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
@@ -99,7 +99,7 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
 extension SOCKSClientHandler {
     
     private func beginHandshake(context: ChannelHandlerContext) {
-        guard context.channel.isActive, self.state.shouldBeginHandshake else {
+        guard context.channel.isActive,  self.state.shouldBeginHandshake else {
             return
         }
         do {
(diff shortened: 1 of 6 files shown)


glbrntt (commented):
> This wasn't caught by the tests as the tests use an EmbeddedChannel, where the channel is always active.

I don't think this is quite right; a newly created `EmbeddedChannel` won't be active until after the `connect` succeeds. I _think_ you can write on an `EmbeddedChannel` which isn't active which might be why the tests didn't pick this up.

The change looks correct but I think we should add a test so we know for sure what we're actually fixing!

Lukasa (changes requested):
I think @glbrntt is right here: we definitely need a test that indicates what's actually happening.

Davidde94:
@glbrntt right you are - my test setup calls connect

Tests/NIOSOCKSTests/SocksClientHandler+Tests.swift
```diff
@@ -152,5 +159,18 @@ class SocksClientHandlerTests: XCTestCase {
             XCTAssertEqual(e as? SOCKSError.ConnectionFailed, .init(reply: .serverFailure))
         }
     }
+    
+    func testDelayedConnection() {
```
glbrntt, line 163: This is just the first part of `testTypicalWorkflow` with an extra assertion, i.e. we're validating the add handler and then connect flow. Should we also add a test for the reverse, i.e. connect and then add the handler?

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -99,7 +99,7 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
 extension SOCKSClientHandler {
     
     private func beginHandshake(context: ChannelHandlerContext) {
-        guard self.state.shouldBeginHandshake else {
+        guard context.channel.isActive,  self.state.shouldBeginHandshake else {
```
glbrntt, line 102:
```suggestion
        guard context.channel.isActive, self.state.shouldBeginHandshake else {
```

Tests/NIOSOCKSTests/ClientRequest+Tests.swift
```diff
@@ -58,13 +58,13 @@ extension ClientRequestTests {
     func testWriteAddressType(){
         var ipv4 = ByteBuffer()
         XCTAssertEqual(ipv4.writeAddressType(.address(try! .init(ipAddress: "192.168.1.1", port: 80))), 7)
-        XCTAssertEqual(ipv4.readBytes(length: 5)!, [1, 1, 1, 168, 192])
-        XCTAssertEqual(ipv4.readInteger(as: UInt16.self)!, 0x5000)
+        XCTAssertEqual(ipv4.readBytes(length: 5)!, [1, 192, 168, 1, 1])
+        XCTAssertEqual(ipv4.readInteger(as: UInt16.self)!, 0x0050)
```
glbrntt, line 62: Might as well make it more obvious:
```suggestion
        XCTAssertEqual(ipv4.readInteger(as: UInt16.self)!, 80)
```

Tests/NIOSOCKSTests/ClientRequest+Tests.swift
```diff
@@ -58,13 +58,13 @@ extension ClientRequestTests {
...
         
         var ipv6 = ByteBuffer()
         XCTAssertEqual(ipv6.writeAddressType(.address(try! .init(ipAddress: "0001:0002:0003:0004:0005:0006:0007:0008", port: 80))), 19)
         XCTAssertEqual(ipv6.readBytes(length: 17)!, [4, 0, 1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 0, 8])
-        XCTAssertEqual(ipv6.readInteger(as: UInt16.self)!, 0x5000)
+        XCTAssertEqual(ipv6.readInteger(as: UInt16.self)!, 0x0050)
```
glbrntt, line 67:
```suggestion
        XCTAssertEqual(ipv6.readInteger(as: UInt16.self)!, 80)
```

glbrntt (approved):
LGTM

Lukasa (approved):

PR #132, opened 2021-06-16
Implement SOCKSv5 server handshake

Davidde94:
RFC 1928

Implement the server side of SOCKSv5. Specifically we've implemented the handshake part of the RFC. Users are expected to add the handler `SOCKSServerHandshakeHandler` to their channel for the handshake to complete.


Davidde94:
Ignore the API breakage - we haven't released the client code yet, all of this will go out in the same release.

Sources/NIOSOCKS/Messages/Messages.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// Contains the proposed authentication methods.
+    case greeting(ClientGreeting)
+    
+    /// Instructs the server of the target host, and the type of connection.
+    case request(SOCKSRequest)
+    
+    /// Data that is sent once the handshake is complete. Handling this case
+    /// should be rare, as you should remove the server handler after the handshake
+    /// process is complete. It can also be used when authenticating in response
+    /// to server challenges.
```
Lukasa, line 29: This latter comment is not quite right: it's used by the server to process unknown auth message types.

Sources/NIOSOCKS/State/ServerStateMachine.swift
```diff
@@ -0,0 +1,142 @@
...
+        try self.guardState(.waitingToSendResponse)
+        if response.reply == .succeeded {
+            self.state = .active
+        } else {
+            self.state = .error
+        }
+    }
+    
+    mutating func sendData() throws {
+        try self.guardState([.authenticating, .active])
```
Lukasa, line 135: Can I propose we use the explicit `switch` statements? They take up more source code space, but they do also force us to examine these areas if we change the state machine.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
```diff
@@ -0,0 +1,191 @@
...
+
+class PromiseTestHandler: ChannelInboundHandler {
+    typealias InboundIn = ClientMessage
+    
+    let expectedGreeting: ClientGreeting
+    let greetingPromise: EventLoopPromise<Void>
+    let expectedRequest: SOCKSRequest
+    let requestPromise: EventLoopPromise<Void>
+    let expectedData: ByteBuffer
+    let dataPromise: EventLoopPromise<Void>
```
Lukasa, line 27: Do these need to be promises, or can they be bools? If we're using embedded channel we presumably don't need the thread safety of promises, and it's much harder to ask the right questions if they're promises.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
```diff
@@ -0,0 +1,191 @@
...
+        self.channel = EmbeddedChannel(handler: self.handler)
+    }
+
+    override func tearDown() {
+        XCTAssertNotNil(self.channel)
+        self.channel = nil
+    }
+    
+    func assertOutputBuffer(_ bytes: [UInt8], line: UInt = #line) {
+        if var buffer = try! self.channel.readOutbound(as: ByteBuffer.self) {
```
Lukasa, line 78: Any reason to make this crash, or should it just throw, or even better XCTFail.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
```diff
@@ -0,0 +1,191 @@
...
+        }
+    }
+    
+    func writeInbound(_ bytes: [UInt8], line: UInt = #line) {
+        try! self.channel.writeInbound(ByteBuffer(bytes: bytes))
+    }
+    
+    func assertInbound(_ bytes: [UInt8], line: UInt = #line) {
+        var buffer = try! self.channel.readInbound(as: ByteBuffer.self)
+        XCTAssertEqual(buffer!.readBytes(length: buffer!.readableBytes), bytes, line: line)
```
Lukasa, line 91: There are a lot of crashes here that would be better manifesting as `XCTFail`s.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
```diff
@@ -0,0 +1,191 @@
...
+    func assertOutputBuffer(_ bytes: [UInt8], line: UInt = #line) {
+        if var buffer = try! self.channel.readOutbound(as: ByteBuffer.self) {
+            XCTAssertEqual(buffer.readBytes(length: buffer.readableBytes), bytes, line: line)
+        } else if bytes.count > 0 {
+            XCTFail("Expected bytes but found none")
+        }
+    }
+    
+    func writeInbound(_ bytes: [UInt8], line: UInt = #line) {
+        try! self.channel.writeInbound(ByteBuffer(bytes: bytes))
```
Lukasa, line 86: Probably should either throw or `XCTFail` instead of crashing.

Lukasa (changes requested):

Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
```diff
@@ -0,0 +1,108 @@
...
+    func handleWriteResponse(
+        _ response: SOCKSResponse, context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
+        var buffer = context.channel.allocator.buffer(capacity: 16)
+        buffer.writeServerResponse(response)
+        try stateMachine.sendServerResponse(response)
+        context.write(self.wrapOutboundOut(buffer), promise: promise)
+    }
+    
+    func handleWriteData(_ data :ByteBuffer, context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
+
```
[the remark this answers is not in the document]
Davidde94: This is why it's a draft PR ;)

Sources/NIOSOCKS/State/ServerStateMachine.swift
```diff
@@ -0,0 +1,142 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+public enum ServerState: Hashable {
```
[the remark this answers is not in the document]
Davidde94: It's contained in the error messages.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
Davidde94: Promises tends to be the NIO way, no? Bools seems nasty for this. If you're against promises then I'd rather use test expectations.

Sources/NIOSOCKS/State/ClientStateMachine.swift
```diff
@@ -14,7 +14,7 @@
 
 import NIO
 
-enum ClientState: Hashable {
+public enum ClientState: Hashable {
```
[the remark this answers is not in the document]
Davidde94: Contained in error messages

Sources/NIOSOCKS/Messages/Messages.swift
```diff
@@ -0,0 +1,66 @@
...
+public enum ClientMessage: Hashable {
+    
+    /// Contains the proposed authentication methods.
+    case greeting(ClientGreeting)
+    
+    /// Instructs the server of the target host, and the type of connection.
+    case request(SOCKSRequest)
+    
+    /// Data that is sent once the handshake is complete. Handling this case
+    /// should be rare, as you should remove the server handler after the handshake
```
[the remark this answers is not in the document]
Davidde94: I've renamed to "authentication data".

Sources/NIOSOCKS/State/ServerStateMachine.swift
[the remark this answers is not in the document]
Lukasa: I don't think we should expose it directly: we can't add new cases to it if we do. I'd rather we stored the state privately and printed it. I don't think users can meaningfully make a decision based on this information anyway.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
Lukasa: Promises are the NIO way when concurrency is involved. Here this is EmbeddedChannel. It's either happened or it hasn't.

Sources/NIOSOCKS/State/ServerStateMachine.swift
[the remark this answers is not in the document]
Davidde94: Yeh I think you're right, I'll remove

Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
```diff
@@ -0,0 +1,120 @@
...
+    var stateMachine: ServerStateMachine
+    
+    public init() {
+        self.stateMachine = ServerStateMachine()
+    }
+    
+    public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        
+        if self.stateMachine.proxyEstablished {
+            context.fireChannelRead(data)
```
Lukasa, line 39: This is incompatible with the way we unbuffer data in `handlerRemoved`: if we get a `channelRead` before `handlerRemoved` fires, we're in trouble. We should probably continue to buffer until we're removed and deliver all the bytes at that point.

Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
```diff
@@ -0,0 +1,120 @@
...
+            try self.stateMachine.sendData()
+            context.write(self.wrapOutboundOut(data), promise: promise)
+        } catch {
+            promise?.fail(error)
+        }
+    }
+    
+    private func handleAuthenticationComplete(context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
+        try stateMachine.authenticationComplete()
+        promise?.succeed(())
```
Lukasa, line 117: We need to issue a zero-length write here. A rule we have about TCP is that sequential write promises must be completed in order, and that requires us to actually issue a write.

Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
```diff
@@ -0,0 +1,120 @@
...
+    
+    private func handleWriteResponse(
+        _ response: SOCKSResponse, context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
+        try stateMachine.sendServerResponse(response)
+        var buffer = context.channel.allocator.buffer(capacity: 16)
+        buffer.writeServerResponse(response)
+        context.write(self.wrapOutboundOut(buffer), promise: promise)
+    }
+    
+    private func handleWriteData(_ data :ByteBuffer, context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
```
Lukasa, line 106:
```suggestion
    private func handleWriteData(_ data: ByteBuffer, context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
```

Sources/NIOSOCKS/Messages/Errors.swift
```diff
@@ -14,68 +14,73 @@
...
 
+/// Wrapper for SOCKS protcol error.
 public enum SOCKSError {
     
-    public struct InvalidState: Error, Hashable {
-        var expected: ClientState
-        var actual: ClientState
+    /// The SOCKS client was in a different state to that required.
+    public struct InvalidClientState: Error, Hashable {
+
```
Lukasa, line 22: Can we add public inits for these two?

Sources/NIOSOCKS/Messages/Messages.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// Sent by the server to inform the client that establishing the proxy to the target
+    /// host succeeded or failed.
+    case response(SOCKSResponse)
+    
+    /// Used when authenticating to send server challenges to the client.
+    case authenticationData(ByteBuffer)
+    
+    /// This is a faux message to update the server's state machine. It should be sent
+    /// once the server is satisified that the client is fully-authenticated.
+    case authenticationComplete
```
Lukasa, line 45: Right now our implementation relies on the server implementer sending this essentially concurrently with their authentication data response. If for any reason it's delayed we'll have trouble, as we'll pass the SOCKS client request message through as though it were an auth message.

It might be better to have `authenticationData` carry a flag that encodes whether this is the last message in the flow. Clients can't use that, but we're doing something a bit different for clients anyway.

Lukasa (changes requested):

Sources/NIOSOCKS/Messages/Messages.swift
Davidde94: Yeh you're right

Lukasa (approved):
LGTM, seems right.

Lukasa:
Merging over the API checker: we haven’t released since we added the APIs being removed.

PR #133, opened 2021-06-16
Fix incorrect SOCKS client flushing behaviour

Davidde94:
Use a `MarkedCircularBuffer` to track pending writes while the SOCKS handshake has not yet completed.

### Motivation:

Data wasn't flushed properly in some cases.

### Modifications:

Use a marked circular buffer to mark on flush, and then write up until that mark when flushing occurs. Also added regression tests.

### DIFF
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
@@ -79,8 +79,11 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
     }
     
     public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
-        let buffer = self.unwrapOutboundIn(data)
-        self.bufferedWrites.append((data, promise))
+        if self.state.proxyEstablished {
+            context.write(data, promise: promise)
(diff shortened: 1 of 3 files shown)


Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -79,21 +79,25 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
...
     
     public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
-        guard self.state.proxyEstablished else {
-            self.bufferedWrites.append((data, promise))
-            return
-        }
-        self.writeBufferedData(context: context)
-        context.write(data, promise: promise)
+        let buffer = self.unwrapOutboundIn(data)
+        self.bufferedWrites.append((data, promise))
```
Lukasa, line 83: This adds overhead into the write path after the negotiation has completed. Rather than unconditionally buffer this, can we keep the fast-path here?

Davidde94: Yeh good idea

Lukasa (approved):
LGTM.

PR #134, opened 2021-06-17
Add SOCKS authentication fast path

Davidde94:
Add a fast path to auto-authenticate when the server selects `noneRequired` as the SOCKS server authentication method.

### DIFF
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
@@ -114,7 +114,7 @@ public final class SOCKSServerHandshakeHandler: ChannelDuplexHandler, RemovableC
             if !self.stateMachine.isAuthenticated {
                 try self.stateMachine.authenticationComplete()
             }
-            promise?.succeed(())
+            context.write(data, promise: promise)
             return
         }
         
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
@@ -69,16 +69,20 @@ public final class SOCKSServerHandshakeHandler: ChannelDuplexHandler, RemovableC
     }
     
     public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+        let message = self.unwrapOutboundIn(data)
+        let outboundBuffer: ByteBuffer
+        
         do {
-            let message = self.unwrapOutboundIn(data)
             switch message {
             case .selectedAuthenticationMethod(let method):
-                try self.handleWriteSelectedAuthenticationMethod(method, context: context, promise: promise)
+                outboundBuffer = try self.handleWriteSelectedAuthenticationMethod(method, context: context)
             case .response(let response):
-                try self.handleWriteResponse(response, context: context, promise: promise)
+                outboundBuffer = try self.handleWriteResponse(response, context: context)
             case .authenticationData(let data, let complete):
-                try self.handleWriteAuthenticationData(data, complete: complete, context: context, promise: promise)
+                outboundBuffer = try self.handleWriteAuthenticationData(data, complete: complete, context: context)
             }
+            context.write(self.wrapOutboundOut(outboundBuffer), promise: promise)
+            
         } catch {
             context.fireErrorCaught(error)
             promise?.fail(error)
@@ -86,24 +90,25 @@ public final class SOCKSServerHandshakeHandler: ChannelDuplexHandler, RemovableC
     }
     
     private func handleWriteSelectedAuthenticationMethod(
-        _ method: SelectedAuthenticationMethod, context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
+        _ method: SelectedAuthenticationMethod, context: ChannelHandlerContext) throws -> ByteBuffer {
         try stateMachine.sendAuthenticationMethod(method)
         var buffer = context.channel.allocator.buffer(capacity: 16)
         buffer.writeMethodSelection(method)
-        context.write(self.wrapOutboundOut(buffer), promise: promise)
+        return buffer
     }
     
     private func handleWriteResponse(
-        _ response: SOCKSResponse, context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
+        _ response: SOCKSResponse, context: ChannelHandlerContext) throws -> ByteBuffer {
         try stateMachine.sendServerResponse(response)
         var buffer = context.channel.allocator.buffer(capacity: 16)
         buffer.writeServerResponse(response)
-        context.write(self.wrapOutboundOut(buffer), promise: promise)
+        return buffer
     }
     
-    private func handleWriteAuthenticationData(_ data: ByteBuffer, complete: Bool, context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
+    private func handleWriteAuthenticationData(
+        _ data: ByteBuffer, complete: Bool, context: ChannelHandlerContext) throws -> ByteBuffer {
         try self.stateMachine.sendAuthenticationData(data, complete: complete)
-        context.write(self.wrapOutboundOut(data), promise: promise)
+        return data
     }
     
 }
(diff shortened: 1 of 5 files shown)


Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
```diff
@@ -102,15 +108,21 @@ public final class SOCKSServerHandshakeHandler: ChannelDuplexHandler, RemovableC
     }
     
     private func handleWriteAuthenticationData(_ data: ByteBuffer, complete: Bool, context: ChannelHandlerContext, promise: EventLoopPromise<Void>?) throws {
-        do {
-            try self.stateMachine.sendData()
-            if complete {
+
```
fabianfett, line 111: Something that I have seen just now. A function that throws and takes a promise seems very odd to me from an API view. Yes this is only a private method... but I think some rules even apply for private methods.

Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
```diff
@@ -91,6 +91,12 @@ public final class SOCKSServerHandshakeHandler: ChannelDuplexHandler, RemovableC
         var buffer = context.channel.allocator.buffer(capacity: 16)
         buffer.writeMethodSelection(method)
         context.write(self.wrapOutboundOut(buffer), promise: promise)
+        
+        // fast path to check if authentication can be marked as complete
+        // only applies when there is no authentication
+        if method.method == .noneRequired {
+            try self.handleWriteAuthenticationData(context.channel.allocator.buffer(capacity: 0), complete: true, context: context, promise: nil)
```
Lukasa, line 98: I think we can reimagine this a bit better: on `sendAuthenticationMethod` with `noneRequired`, the state machine should simply transition into the expected state. We also need to keep track of the auth method we used so that we can ensure we tolerate zero-sized auth data with complete if we used the right method.

Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
```diff
@@ -102,15 +108,21 @@ public final class SOCKSServerHandshakeHandler: ChannelDuplexHandler, RemovableC
...
+        
+        // fast path to check if we can automatically mark authentication as complete
+        if data.readableBytes == 0 && complete {
+            if !self.stateMachine.isAuthenticated {
                 try self.stateMachine.authenticationComplete()
             }
-            context.write(self.wrapOutboundOut(data), promise: promise)
-        } catch {
-            promise?.fail(error)
+            promise?.succeed(())
```
Lukasa, line 117: We do still have to issue the zero-length write here.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
```diff
@@ -229,4 +229,40 @@ class SOCKSServerHandlerTests: XCTestCase {
...
+        self.writeInbound([0x05, 0x01, 0x00])
+        self.writeOutbound(.selectedAuthenticationMethod(.init(method: .noneRequired)))
+        self.assertOutputBuffer([0x05, 0x00])
+        
+        // complete authentication, but nothing should be written
+        // to the network
+        self.writeOutbound(.authenticationData(ByteBuffer(), complete: true))
+        self.assertOutputBuffer([])
+        
+        // if we try and write the request then it should fail *if* auth hasn't
```
Lukasa, line 262: Can we validate this comment (i.e. prove that it actually does fail?)

Lukasa (commented):
Cool, generally looks good, some notes inline.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
Davidde94: Upon re-reading this comment it's actually a lie. I've updated to reflect what will actually happen. It's also tested above, but I've added a test explicitly to test auth too.

Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
Davidde94: I've refactored, see what you think :)

fabianfett: Looks good!

Sources/NIOSOCKS/Channel Handlers/SOCKSServerHandshakeHandler.swift
```diff
@@ -69,48 +69,46 @@ public final class SOCKSServerHandshakeHandler: ChannelDuplexHandler, RemovableC
     }
     
     public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+        let message = self.unwrapOutboundIn(data)
+        let outboundBuffer: ByteBuffer
```
Lukasa, line 73: Nit: this variable is used entirely within the scope of `do`, and so can be defined there.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
```diff
@@ -229,4 +229,60 @@ class SOCKSServerHandlerTests: XCTestCase {
         // removing the handler, it should fail
         XCTAssertThrowsError(try self.channel.writeOutbound(ServerMessage.authenticationData(ByteBuffer(string: "hello, world!"), complete: false)))
     }
+    
+    func testAutoAuthentictionComplete() {
```
Lukasa, line 233:
```suggestion
    func testAutoAuthenticationComplete() {
```

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
```diff
@@ -229,4 +229,60 @@ class SOCKSServerHandlerTests: XCTestCase {
...
+        
+        // if we try and write the request then the data would be read
+        // as authentication data, and so the server wouldn't reply
+        // with a response
+        self.writeInbound([0x05, 0x01, 0x00, 0x01, 127, 0, 0, 1, 0, 80])
+        self.writeOutbound(.response(.init(reply: .succeeded, boundAddress: .address(try! .init(ipAddress: "127.0.0.1", port: 80)))))
+        self.assertOutputBuffer([0x05, 0x00, 0x00, 0x01, 127, 0, 0, 1, 0, 80])
+    }
+    
+    func testAutoAuthentictionCompleteWithManuallyCompletion() {
```
Lukasa, line 249:
```suggestion
    func testAutoAuthenticationCompleteWithManualCompletion() {
```

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
```diff
@@ -229,4 +229,60 @@ class SOCKSServerHandlerTests: XCTestCase {
...
+        // server selects none-required, this should mean we can continue without
+        // having to manually inform the state machine. However, informing the state
+        // machine manually shouldn't break anything.
+        self.writeInbound([0x05, 0x01, 0x00])
+        self.writeOutbound(.selectedAuthenticationMethod(.init(method: .noneRequired)))
+        self.assertOutputBuffer([0x05, 0x00])
+        
+        // complete authentication, but nothing should be written
+        // to the network
+        self.writeOutbound(.authenticationData(ByteBuffer(), complete: true))
```
Lukasa, line 260: Can we add two more tests:

1. Validate that this write fails if there is a non-zero number of bytes here
2. Validate that this write fails if `complete` is `false`.

Lukasa (changes requested):
A few nits, and then I'd love to see two more tests just to cover our bases, then we're good.

Tests/NIOSOCKSTests/SOCKSServerHandshakeHandler+Tests.swift
```diff
@@ -229,4 +229,88 @@ class SOCKSServerHandlerTests: XCTestCase {
...
+    func testManualAuthenticationFailureExtraBytes() {
+        // server selects none-required, this should mean we can continue without
+        // having to manually inform the state machine. However, informing the state
+        // machine manually shouldn't break anything.
+        self.writeInbound([0x05, 0x01, 0x00])
+        self.writeOutbound(.selectedAuthenticationMethod(.init(method: .noneRequired)))
+        self.assertOutputBuffer([0x05, 0x00])
+        
+        // invalid authentication completion
+        // we've selection `noneRequired`, so no
```
Lukasa, line 297:
```suggestion
        // we've selected `noneRequired`, so no
```

Lukasa (approved):
Great thanks. Clean up the typo and we're good to go.

PR #136, opened 2021-06-18
Add SOCKSProxyEstablishedEvent

fabianfett:
Adds a `SOCKSProxyEstablishedEvent` that is issued once the SOCKS connection is established.

### Motivation:

- We want to know when the SOCKS proxy connection was established.

### Modifications:

- The `SOCKSClientHandler` and `SOCKSServerHandshakeHandler` issue a connect event

### Result:

- We can observe successful proxy connection handshakes

### DIFF
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
@@ -74,7 +74,7 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
             try self.handleAction(action, context: context)
         } catch {
             context.fireErrorCaught(error)
-            context.close(mode: .all, promise: nil)
+            context.close(promise: nil)
         }
     }
     
@@ -151,6 +151,7 @@ extension SOCKSClientHandler {
         // If we have any buffered writes then now
         // we can send them.
         self.writeBufferedData(context: context)
+        context.fireUserInboundEventTriggered(SOCKSProxyEstablishedEvent())
     }
     
     private func handleActionSendRequest(context: ChannelHandlerContext) throws {
@@ -165,4 +166,23 @@ extension SOCKSClientHandler {
         context.writeAndFlush(self.wrapOutboundOut(buffer), promise: nil)
     }
     
+}
+
+extension SOCKSClientHandler: RemovableChannelHandler {
+    
+    public func removeHandler(context: ChannelHandlerContext, removalToken: ChannelHandlerContext.RemovalToken) {
+        guard self.state.proxyEstablished else {
+            preconditionFailure("The SOCKSClientHandler can only be removed once a connection has been established")
+        }
+        context.leavePipeline(removalToken: removalToken)
+    }
+    
+}
+
+/// A `Channel` user event that is sent when a SOCKS connection has been established
+///
+/// After this event has been received it is save to remove the `SOCKSClientHandler` from the channel pipeline.
+public struct SOCKSProxyEstablishedEvent {
+    public init() {
+    }
 }
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
@@ -28,6 +28,7 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
     private let targetAddress: SOCKSAddress
     
     private var state: ClientStateMachine
+    private var removalToken: ChannelHandlerContext.RemovalToken?
     private var inboundBuffer: ByteBuffer?
     
     private var bufferedWrites: MarkedCircularBuffer<(NIOAny, EventLoopPromise<Void>?)> = .init(initialCapacity: 8)
@@ -95,6 +96,11 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
             context.write(data, promise: promise)
         }
         context.flush() // safe to flush otherwise we wouldn't have the mark
+        
+        while !self.bufferedWrites.isEmpty {
+            let (data, promise) = self.bufferedWrites.removeFirst()
+            context.write(data, promise: promise)
+        }
     }
     
     public func flush(context: ChannelHandlerContext) {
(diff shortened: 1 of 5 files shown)


Davidde94 (approved):

Lukasa:
Let's go with #137.

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -151,6 +151,7 @@ extension SOCKSClientHandler {
         // If we have any buffered writes then now
         // we can send them.
         self.writeBufferedData(context: context)
+        context.fireUserInboundEventTriggered(SOCKSProxyEstablishedEvent())
```
Lukasa, line 154: Can I propose that we fire this _first_? Otherwise we will be delivering application bytes before we've fired this user event, which makes managing things very hard.

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -166,3 +167,22 @@ extension SOCKSClientHandler {
     }
     
 }
+
+extension SOCKSClientHandler: RemovableChannelHandler {
+    
+    public func removeHandler(context: ChannelHandlerContext, removalToken: ChannelHandlerContext.RemovalToken) {
+        guard self.state.proxyEstablished else {
+            preconditionFailure("The SOCKSClientHandler can only be removed once a connection has been established")
+        }
```
fabianfett, line 176: We could work around this precondition by adding this to the state machine. Thus we would be waiting for the SOCKS handshake to succeed or fail and only trigger the leave then. @Davidde94 @Lukasa is this something we want to add?

fabianfett:
Changed to needs minor version bump since we conform `SOCKSClientHandler` to `RemovableChannelHandler`.

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
Davidde94: Adding what to the state machine? A new action "remove handler"?

Lukasa: Yeah, I think that works.

Lukasa (changes requested):

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
fabianfett: If we fire this event first, we might remove the ChannelHandler before all data is written... I think with this, we def. need to add the wanted removal to the state machine.

Lukasa: It's ok to remove the channel handler first. There are two important notes to channel handler removal:

1. No-one can force a handler out of the channel before it's willing to go
2. During `handlerRemoved` you still have a context that can be used to forward data.

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -165,4 +162,42 @@ extension SOCKSClientHandler {
         context.writeAndFlush(self.wrapOutboundOut(buffer), promise: nil)
     }
     
+    private func emptyInboundAndOutboundBuffer(context: ChannelHandlerContext) {
+        if let inboundBuffer = self.inboundBuffer, inboundBuffer.readableBytes > 0 {
+            // after the SOCKS handshake message we already received further bytes.
+            // so let's send them down the pipe
+            let data = self.wrapInboundOut(self.inboundBuffer!)
```
Lukasa, line 169: We don't need a force unwrap here, we have done a conditional binding. Also, we can move this wrap into the call to `fireChannelRead`.

fabianfett: Urgh. Oversight. Fixed.

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
Davidde94: Discussed offline - I think that the state machine should be completely decoupled from channel handler logic. The proposed solution was to store an optional removal token in the handler, and deal with removal when the state machine emits `proxy established`.

Tests/NIOSOCKSTests/SocksClientHandler+Tests.swift
```diff
@@ -230,5 +236,110 @@ class SocksClientHandlerTests: XCTestCase {
...
+            }
+        }
+        
+        let establishPromise = self.channel.eventLoop.makePromise(of: Void.self)
+        let removalPromise = self.channel.eventLoop.makePromise(of: Void.self)
+        establishPromise.futureResult.whenSuccess { _ in
+            self.channel.pipeline.removeHandler(self.handler).cascade(to: removalPromise)
+        }
+        
+        try! self.channel.pipeline.addHandler(SOCKSEventHandler(establishedPromise: establishPromise)).wait()
```
Davidde94, line 267: Best to wrap in an `XCTAssertNoThrow` so we can lose the bang?

Tests/NIOSOCKSTests/SocksClientHandler+Tests.swift
```diff
@@ -230,5 +236,110 @@ class SocksClientHandlerTests: XCTestCase {
...
+        
+        self.connect()
+        
+        self.channel.write(ByteBuffer(bytes: [1, 2, 3]), promise: nil)
```
Davidde94, line 271: Could you just add a comment saying that these writes should be correctly buffered. Also to simplify the test you can probably reduce the amount of buffered data and the flush. This is tested elsewhere.

Tests/NIOSOCKSTests/SocksClientHandler+Tests.swift
```diff
@@ -230,5 +236,110 @@ class SocksClientHandlerTests: XCTestCase {
...
+        XCTAssertNoThrow(try removalPromise.futureResult.wait())
+        XCTAssertThrowsError(try self.channel.pipeline.syncOperations.handler(type: SOCKSClientHandler.self)) {
+            XCTAssertEqual($0 as? ChannelPipelineError, .notFound)
+        }
+    }
+    
+    func testHandlerRemovalBeforeConnectionIsEstablished() {
+        self.connect()
+        
+        self.channel.write(ByteBuffer(bytes: [1, 2, 3]), promise: nil)
```
Davidde94, line 296: Same comment re buffered data

Davidde94 (approved):
Looks good, a few nits

Lukasa (approved):

PR #137, opened 2021-06-18
SOCKSClientHandler removes itself on proxy established

fabianfett:
Alternative to #136. 

### Motivation:

- Once a SOCKS connection has been established we want to remove the `SOCKSClientHandler` from the channel pipeline and potentially do other pipeline modifications.

### Modifications:

- The `SOCKSClientHandler` removes itself from the channel pipeline once the connection was successfully established.
- An `establishHandler: @escaping ((Channel) -> EventLoopFuture<Void>)` is offered to allow for other pipeline modifications.

### Result:

- We can observe successful proxy connection handshakes and the user does not have to remove the `SOCKSClientHandler` themselves

### DIFF
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
@@ -166,7 +166,7 @@ extension SOCKSClientHandler {
         // If we have any buffered writes then now we can send them.
         self.writeBufferedData(context: context)
         
-        // Remove ourselfes from the channel pipeline and fulfill the establishPromise
+        // Remove ourselves from the channel pipeline and fulfill the establishPromise
         context.channel.pipeline.removeHandler(self).cascade(to: self.connectionEstablishedPromise)
     }
     
--- a/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
+++ b/Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
@@ -16,9 +16,11 @@ import NIO
 
 /// Connects to a SOCKS server to establish a proxied connection
 /// to a host. This handler should be inserted at the beginning of a
-/// channel's pipeline. Note that SOCKS only supports fully-qualified
-/// domain names and IPv4 or IPv6 sockets, and not UNIX sockets.
-public final class SOCKSClientHandler: ChannelDuplexHandler {
+/// channel's pipeline. It will remove itself once the SOCKS proxy
+/// connection has been established. Note that SOCKS only supports
+/// fully-qualified domain names and IPv4 or IPv6 sockets, and
+/// not UNIX sockets.
+public final class SOCKSClientHandler: ChannelDuplexHandler, RemovableChannelHandler {
     
     public typealias InboundIn = ByteBuffer
     public typealias InboundOut = ByteBuffer
@@ -31,11 +33,20 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
     private var inboundBuffer: ByteBuffer?
     
     private var bufferedWrites: MarkedCircularBuffer<(NIOAny, EventLoopPromise<Void>?)> = .init(initialCapacity: 8)
+    private let connectionEstablishedPromise: EventLoopPromise<Void>?
     
     /// Creates a new `SOCKSClientHandler` that connects to a server
     /// and instructs the server to connect to `targetAddress`.
     /// - parameter targetAddress: The desired end point - note that only IPv4, IPv6, and FQDNs are supported.
-    public init(targetAddress: SOCKSAddress) {
+    public convenience init(targetAddress: SOCKSAddress) {
+        self.init(targetAddress: targetAddress, connectionEstablishedPromise: nil)
+    }
+    
+    /// Creates a new `SOCKSClientHandler` that connects to a server
+    /// and instructs the server to connect to `targetAddress`.
+    /// - parameter targetAddress: The desired end point - note that only IPv4, IPv6, and FQDNs are supported.
+    /// - parameter connectionEstablishedPromise: An optional promise that is succeeded once the SOCKS connection is established.
+    public init(targetAddress: SOCKSAddress, connectionEstablishedPromise: EventLoopPromise<Void>?) {
         
         switch targetAddress {
         case .address(.unixDomainSocket):
@@ -46,6 +57,7 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
         
         self.state = ClientStateMachine()
         self.targetAddress = targetAddress
+        self.connectionEstablishedPromise = connectionEstablishedPromise
     }
     
     public func channelActive(context: ChannelHandlerContext) {
@@ -73,8 +85,7 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
             let action = try self.state.receiveBuffer(&self.inboundBuffer!)
             try self.handleAction(action, context: context)
         } catch {
-            context.fireErrorCaught(error)
-            context.close(mode: .all, promise: nil)
+            self.handleError(error, context: context)
         }
     }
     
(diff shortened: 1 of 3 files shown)


Davidde94 (commented):
Can we add a test for if the future in the callback is failed?

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -31,11 +31,16 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
...
     
     private var bufferedWrites: MarkedCircularBuffer<(NIOAny, EventLoopPromise<Void>?)> = .init(initialCapacity: 8)
+    private var establishHandler: (Channel) -> EventLoopFuture<Void>
     
     /// Creates a new `SOCKSClientHandler` that connects to a server
     /// and instructs the server to connect to `targetAddress`.
     /// - parameter targetAddress: The desired end point - note that only IPv4, IPv6, and FQDNs are supported.
-    public init(targetAddress: SOCKSAddress) {
+    public convenience init(targetAddress: SOCKSAddress) {
+        self.init(targetAddress: targetAddress, establishHandler: { $0.eventLoop.makeSucceededFuture(()) })
```
Lukasa, line 40:
```suggestion
        self.init(targetAddress: targetAddress, establishHandler: { $0.eventLoop.makeSucceededVoidFuture(()) })
```

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -31,11 +31,16 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
...
+    }
+    
+    public init(targetAddress: SOCKSAddress, establishHandler: @escaping ((Channel) -> EventLoopFuture<Void>)) {
```
Lukasa, line 43: Per David's note, it would be good to establish that this actually does delay removal.

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
fabianfett: Is there any reasoning behind this, besides personal preference? I would assume that the compiler always uses `makeSucceededVoidFuture` given: https://github.com/apple/swift-nio/blob/main/Sources/NIO/EventLoop.swift#L602-L605 Personally I like to construct all my futures in exactly the same way, and assume that the compiler does the right thing.

And don't get me wrong happy to change. :) Just want to understand the reasoning...

Lukasa: It's personal preference. The advantage of calling the specialized method is that you always call it: you don't rely on the optimiser seeing what you're doing and making the call go away. That's it. :)

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -31,11 +31,16 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
     private var inboundBuffer: ByteBuffer?
     
     private var bufferedWrites: MarkedCircularBuffer<(NIOAny, EventLoopPromise<Void>?)> = .init(initialCapacity: 8)
+    private let establishHandler: (Channel) -> EventLoopFuture<Void>
```
Davidde94, line 34: Major nit: I'd probably rename to `establishedHandler` or `connectionEstablishedHandler`. It's a little more clear IMHO.

Tests/NIOSOCKSTests/SocksClientHandler+Tests.swift
```diff
@@ -230,5 +235,48 @@ class SocksClientHandlerTests: XCTestCase {
         XCTAssertNoThrow(self.channel.pipeline.addHandler(handler))
         self.assertOutputBuffer([0x05, 0x01, 0x00])
     }
+    
+    func testDelayedHandlerRemoved() {
+        // for this test we can't use the created channel and handler by setUp
+        self.channel = EmbeddedChannel()
+        let promise = channel.eventLoop.makePromise(of: Void.self)
```
Davidde94, line 242:
```suggestion
        let promise = self.channel.eventLoop.makePromise(of: Void.self)
```

Tests/NIOSOCKSTests/SocksClientHandler+Tests.swift
```diff
@@ -230,5 +235,48 @@ class SocksClientHandlerTests: XCTestCase {
...
+        self.handler = SOCKSClientHandler(targetAddress: .address(try! .init(ipAddress: "192.168.1.1", port: 80))) {
+            channel -> EventLoopFuture<Void> in
+
```
Davidde94, line 245:
```suggestion
            
```

Tests/NIOSOCKSTests/SocksClientHandler+Tests.swift
```diff
@@ -230,5 +235,48 @@ class SocksClientHandlerTests: XCTestCase {
...
+        self.writeInbound([0x05, 0x00])
+        
+        // client sends the request
+        self.assertOutputBuffer([0x05, 0x01, 0x00, 0x01, 192, 168, 1, 1, 0x00, 0x50])
+        
+        // server replies yay
+        self.writeInbound([0x05, 0x00, 0x00, 0x01, 192, 168, 1, 1, 0x00, 0x50])
+        
+        // the SOCKS handler remains in the channel pipeline until we succeed or fail the promise
+        XCTAssertNoThrow(try self.channel.pipeline.syncOperations.handler(type: SOCKSClientHandler.self))
```
Davidde94, line 265: Can we check that the data still goes straight through the handler before it's removed? That's very important, otherwise we might lose bytes!

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -31,11 +33,16 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
     private var inboundBuffer: ByteBuffer?
     
     private var bufferedWrites: MarkedCircularBuffer<(NIOAny, EventLoopPromise<Void>?)> = .init(initialCapacity: 8)
+    private let establishPromise: EventLoopPromise<Void>?
```
Davidde94, line 36: I still dislike `establishPromise` haha

Davidde94 (approved):

Tests/NIOSOCKSTests/SocksClientHandler+Tests.swift
fabianfett: Not applicable anymore

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -148,9 +159,11 @@ extension SOCKSClientHandler {
             context.fireChannelRead(data)
         }
         
-        // If we have any buffered writes then now
-        // we can send them.
+        // If we have any buffered writes then now we can send them.
         self.writeBufferedData(context: context)
+        
+        // Remove ourselfes from the channel pipeline and fulfill the establishPromise
```
Lukasa, line 169: Nit:

```suggestion
        // Remove ourselves from the channel pipeline and fulfill the establishPromise
```

Lukasa (approved):

Sources/NIOSOCKS/Channel Handlers/SOCKSClientHandler.swift
```diff
@@ -31,11 +33,20 @@ public final class SOCKSClientHandler: ChannelDuplexHandler {
...
-    public init(targetAddress: SOCKSAddress) {
+    public convenience init(targetAddress: SOCKSAddress) {
+        self.init(targetAddress: targetAddress, connectionEstablishedPromise: nil)
+    }
+    
+    /// Creates a new `SOCKSClientHandler` that connects to a server
+    /// and instructs the server to connect to `targetAddress`.
+    /// - parameter targetAddress: The desired end point - note that only IPv4, IPv6, and FQDNs are supported.
+    /// - parameter connectionEstablishedPromise: An optional promise that is succeeded once the SOCKS connection is established.
+    public init(targetAddress: SOCKSAddress, connectionEstablishedPromise: EventLoopPromise<Void>?) {
```
fabianfett, line 49: Having thought about this for even longer, I'm not sure if this is the best solution (given that this shall become public API). The problem at hand is that using this initializer becomes very hard in `.channelInitializer`, since they might be invoked multiple times. Of course we could make a `var connectionEstablishedFuture: EventLoopFuture<Void>?` public, but using this API would be quite bad.

PR #141, opened 2021-09-01
pin jazzy version

tomerd:
motivation: fix ci

changes: pin jazzy version as newer ones do not work well with older ruby versions


fabianfett (approved):

tomerd:
looks like jazzy and other ruby related dependencies are still very much a moving target. maybe we should only install these on the latest ubuntu version, and then generate docs using that?

PR #155, opened 2022-04-21
add the NFS3 protocol

weissi:
### Motivation:

NFS can be a very useful protocol.

### Modifications:

Add the (read side, for now) of the NFS3 protocol.
This adds (partial) implementations of the following RFCs:

- [RFC 1813: NFS Version 3 Protocol](https://datatracker.ietf.org/doc/html/rfc1813)
- [RFC 1057: RPC: Remote Procedure Call Protocol](https://datatracker.ietf.org/doc/html/rfc1057)
- [RFC 1014: XDR: External Data Representation](https://datatracker.ietf.org/doc/html/rfc1014)

Essentially, NFS3 is a protocol which sends metadata & does framing via the (SUN) RPC protocol. The data is encoded using XDR. If you're familiar with gRPC, then XDR is NFS's protobuf, HTTP is NFS's RPC and NFS3 is the application protocol.

### Result:

- More supported protocols

### Left to do before merging:

- [x] Look at the TODOs
- [x] Discuss public API changes if more of the NFS protocol is implemented
- [x] Rename the module to `_NIONFS3Preview`?
- [x] Decide if the dependency on `https://github.com/apple/swift-distributed-tracing-baggage.git` is acceptable


Package.swift
```diff
@@ -120,6 +135,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.32.0"),
+        .package(url: "https://github.com/apple/swift-distributed-tracing-baggage", .upToNextMajor(from: "0.2.1")),
```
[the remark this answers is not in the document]
weissi: it's only available for pre-1.0 (and I think won't be for a little while, CC @ktoso). The only way to not get this dependency is if we don't actually ship the `NFS3FileSystemHandler` (which of course is an option).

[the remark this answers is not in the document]
weissi: @fabianfett removed

Package.swift
```diff
@@ -108,7 +108,21 @@ var targets: [PackageDescription.Target] = [
             "NIOSOCKS",
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOEmbedded", package: "swift-nio"),
-        ])
+        ]),
+    .target(
+        name: "NIONFS3",
+        dependencies: [
+            .product(name: "NIO", package: "swift-nio"),
```
fabianfett, line 115: Why depend on umbrella NIO? IIRC we want to deprecate importing the umbrella module?

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,526 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+public struct RPCNFSProcedureID: Equatable {
+    public internal(set) var program: UInt32
+    public internal(set) var procedure: UInt32
+
+    public static let mountNull: Self = .init(program: 100005, procedure: 0)
```
glbrntt, line 21: Would be nice to have a link to the source of truth for these values

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,526 @@
...
+    case fsstat(NFS3CallFSStat)
+    case access(NFS3CallAccess)
+    case lookup(NFS3CallLookup)
+    case readdirplus(NFS3CallReadDirPlus)
+    case read(NFS3CallRead)
+    case readlink(NFS3CallReadlink)
+    case setattr(NFS3CallSetattr)
+    case readdir(NFS3CallReadDir)
+
+    case _PLEASE_DO_NOT_EXHAUSTIVELY_MATCH_THIS_ENUM_NEW_CASES_MIGHT_BE_ADDED_IN_THE_FUTURE
```
glbrntt, line 112: Wonderful

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,526 @@
...
+}
+
+internal func nfsStringFillBytes(_ byteCount: Int) -> Int {
+    return (4 - (byteCount % 4)) % 4
+}
+
+extension ByteBuffer {
+    mutating func readRPCVerifier() throws -> RPCOpaqueAuth {
+        guard let flavor = self.readInteger(endianness: .big, as: UInt32.self),
+              let length = self.readInteger(endianness: .big, as: UInt32.self) else {
```
glbrntt, line 155: `readMultipleIntegers`?

Package.swift
weissi: @fabianfett Honestly, I think that's the right thing to do because given how SwiftPM works it's just too easy to forget a dependency and not be told about it. It also feels wrong to burden normal users with even more NIO imports. But I'll follow the remaining code here and add the NIO{Embedded,Core,Posix,...} imports.

Sources/NIONFS3/NFSTypes+Mount.swift
```diff
@@ -0,0 +1,82 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+// MARK: - Mount
+public struct NFS3CallMount: Equatable {
```
glbrntt, line 18: These types all appear to have `Mount` counterparts; are both types used and/or needed?

glbrntt (commented):
I left a bunch of comments inline. I have no doubt that I missed things too.

A few general notes: 
- There are a bunch of message types which are only `Equatable` -- these should probably be `Hashable` and `Sendable` too.
- The `ByteBuffer` read helpers don't follow the normal pattern of returning `nil` if there aren't enough bytes and the write helpers don't return the number of bytes written.
- The Mount* types seem to be duplicated as NFS* types. The latter don't appear to be used?

Sources/NIONFS3/MountTypes+Mount.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public mutating func writeNFSReplyMount(_ reply: MountReplyMount) {
+        self.writeNFSResultStatus(reply.result)
+
+        switch reply.result {
+        case .okay(let reply):
+            self.writeNFSFileHandle(reply.fileHandle)
+            precondition(reply.authFlavors == [.unix] || reply.authFlavors == [.noAuth],
+                         "Sorry, anything but [.unix] / [.system] / [.noAuth] unimplemented.")
```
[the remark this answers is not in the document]
weissi: `.unix` == `.system`, that aliasing is also the same in the spec. I think it was "system" first (because UNIX was always "the system") but then Windows and friends added NFS support and I reckon the RFC process suggested to  make `.unix` the canonical one?

Sources/NIONFS3/MountTypes+Mount.swift
```diff
@@ -0,0 +1,82 @@
...
+    public mutating func readNFSReplyMount() throws -> MountReplyMount {
+        return MountReplyMount(result: try self.readNFSResult(readOkay: { buffer in
+            let fileHandle = try buffer.readNFSFileHandle()
+            let authFlavors = try buffer.readNFSList(readEntry: { buffer in
+                try buffer.readRPCAuthFlavor()
+            })
+            return .init(fileHandle: fileHandle, authFlavors: authFlavors)
+
+        },
+                                                            readFail: { _ in NFS3Nothing() }))
```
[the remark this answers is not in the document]
weissi: reformatted that, thanks.

Sources/NIONFS3/MountTypes+Null.swift
```diff
@@ -0,0 +1,29 @@
...
+
+import NIOCore
+
+// MARK: - Null
+public struct MountCallNull: Equatable {
+    public init() {}
+}
+
+extension ByteBuffer {
+    public mutating func readMountCallNull() throws -> MountCallNull {
```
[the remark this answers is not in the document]
weissi: but conceptually we are so I think this is correct.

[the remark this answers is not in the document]
weissi: you don't want to leak to the caller that the MOUNT call in the MOUNT protocol has no payload bytes.

Sources/NIONFS3/MountTypes+Unmount.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public mutating func writeNFSCallUnmount(_ call: MountCallUnmount) {
+        self.writeNFSString(call.dirPath)
+    }
+
+    public mutating func writeNFSReplyUnmount(_ reply: MountReplyUnmount) {
+    }
+
+    public mutating func readNFSReplyUnmount() throws -> MountReplyUnmount {
```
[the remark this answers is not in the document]
weissi: same here, conceptually this can fail. It just happens to have 0 bytes in the current protocol version.

Sources/NIONFS3/NFSFileSystem.swift
```diff
@@ -0,0 +1,68 @@
...
+extension NFS3FileSystemNoAuth {
+    public func readdir(_ call: NFS3CallReadDir, promise originalPromise: EventLoopPromise<NFS3ReplyReadDir>) {
+        let promise = originalPromise.futureResult.eventLoop.makePromise(of: NFS3ReplyReadDirPlus.self)
+        self.readdirplus(NFS3CallReadDirPlus(fileHandle: call.fileHandle,
+                                             cookie: call.cookie,
+                                             cookieVerifier: call.cookieVerifier,
+                                             dirCount: .max,
+                                             maxCount: call.maxResultByteCount),
+
+                         promise: promise)
```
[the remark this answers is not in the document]
weissi: ta

Sources/NIONFS3/NFSFileSystemHandler.swift
```diff
@@ -0,0 +1,95 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+
+/// `ChannelHandler` which implements NFS calls & replies the user implements as a `NFS3FileSystemNoAuth`.
+///
+/// `NFS3FileSystemNoAuthHandler` is a all-in-one SwiftNIO `ChannelHandler` that implements an NFS3 server. Every call
```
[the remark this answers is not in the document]
weissi: ta

Sources/NIONFS3/NFSFileSystemHandler.swift
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    func sendError(_ error: Error, call: RPCNFS3Call) {
+        if let context = self.context {
+            context.fireErrorCaught(error)
+            context.writeAndFlush(self.wrapOutboundOut(.init(rpcReply: .init(xid: call.rpcCall.xid,
+                                                                             status: self.rpcReplySuccess),
+                                                             nfsReply: .mount(.init(result: .fail(.errorSERVERFAULT,
+                                                                                                  NFS3Nothing()))))),
```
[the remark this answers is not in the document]
weissi: fixed

Sources/NIONFS3/NFSFileSystemHandler.swift
```diff
@@ -0,0 +1,95 @@
...
+                                  promise: nil)
+        }
+    }
+
+
+    public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let call = self.unwrapInboundIn(data)
+        self.invoker!.handleNFSCall(call)
```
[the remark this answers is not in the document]
weissi: done

Sources/NIONFS3/NFSReplyDecoder.swift
```diff
@@ -0,0 +1,61 @@
...
+import NIOCore
+
+public struct NFS3ReplyDecoder: WriteObservingByteToMessageDecoder {
+    public typealias OutboundIn = RPCNFS3Call
+    public typealias InboundOut = RPCNFS3Reply
+
+    private var procedures: [UInt32: RPCNFSProcedureID]
+    private let allowDuplicateReplies: Bool
+
+    public init(prepopulatedProcecedures: [UInt32: RPCNFSProcedureID]? = nil,
```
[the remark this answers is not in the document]
weissi: If you default them to `[:]` then it becomes API that is `[:]`. With `nil` we make that an impl detail.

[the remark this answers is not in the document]
weissi: @glbrntt Mostly testing. Maybe in theory we could do some resumption stuff... But I'll document it.

Sources/NIONFS3/NFSReplyDecoder.swift
```diff
@@ -0,0 +1,61 @@
...
+            progAndProc = p
+        }
+
+        let nfsReply = try body.readNFSReply(programAndProcedure: progAndProc, rpcReply: reply)
+        context.fireChannelRead(self.wrapInboundOut(nfsReply))
+        return .continue
+    }
+
+    public mutating func write(data: RPCNFS3Call) {
+        self.procedures[data.rpcCall.xid] = data.rpcCall.programAndProcedure
```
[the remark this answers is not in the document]
weissi: Not really.

Sources/NIONFS3/NFSTypes+Common.swift
```diff
@@ -0,0 +1,600 @@
...
+
+    public var id: ID {
+        return self.rpcReply.xid
+    }
+}
+
+public enum NFS3Result<Okay, Fail> {
+    case okay(Okay)
+    case fail(NFS3Status, Fail)
+}
```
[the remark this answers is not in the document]
weissi: I think it makes sense to have this specific type for NFS. They're all successes on the protocol layer, it's an application level concern. So I don't think `Swift.Result` makes too  much sense here.

Sources/NIONFS3/NFSTypes+Common.swift
```diff
@@ -0,0 +1,600 @@
...
+public typealias NFS3FileMode = UInt32
+public typealias NFS3UID = UInt32
+public typealias NFS3GID = UInt32
+public typealias NFS3Size = UInt64
+public typealias NFS3SpecData = UInt64
+public typealias NFS3FileID = UInt64
+public typealias NFS3Cookie = UInt64
+public typealias NFS3CookieVerifier = UInt64
+public typealias NFS3Offset = UInt64
+public typealias NFS3Count = UInt32
```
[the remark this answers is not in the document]
weissi: I thought this is the right tradeoff w.r.t. convenience and docs, especially when reading the RFC. The RFC defines that say `size3` (meaning NFSv3's `size`) is a `UInt64` but subsequently talks about `size3` only. So I thought this is kinda handy to have like this. But we can also make it all opaque and create structs for those.

Sources/NIONFS3/NFSTypes+Common.swift
```diff
@@ -0,0 +1,600 @@
...
+
+public struct NFS3Nothing: Equatable {
+    public init() {}
+}
+
+public enum NFS3Status: UInt32 {
+   case ok = 0
```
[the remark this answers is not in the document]
weissi: this is afaik exhaustive

Sources/NIONFS3/NFSTypes+Common.swift
```diff
@@ -0,0 +1,600 @@
...
+   case errorSTALE       = 70
+   case errorREMOTE      = 71
+   case errorBADHANDLE   = 10001
+   case errorNOT_SYNC    = 10002
+   case errorBAD_COOKIE  = 10003
+   case errorNOTSUPP     = 10004
+   case errorTOOSMALL    = 10005
+   case errorSERVERFAULT = 10006
+   case errorBADTYPE     = 10007
+   case errorJUKEBOX     = 10008
```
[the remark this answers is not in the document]
weissi: yeah jukebox is just 'try again please'. We actually use this internally :)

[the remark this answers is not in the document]
weissi: I'll add links

Sources/NIONFS3/NFSTypes+Common.swift
```diff
@@ -0,0 +1,600 @@
...
+
+    public static let read: NFS3Access = .init(rawValue: 0x0001)
+    public static let lookup: NFS3Access = .init(rawValue: 0x0002)
+    public static let modify: NFS3Access = .init(rawValue: 0x0004)
+    public static let extend: NFS3Access = .init(rawValue: 0x0008)
+    public static let delete: NFS3Access = .init(rawValue: 0x0010)
+    public static let execute: NFS3Access = .init(rawValue: 0x0020)
+
+    public static let all: NFS3Access = [.read, .lookup, .modify, .extend, .delete, .execute]
+    public static let allReadOnly: NFS3Access = [.read, .lookup, .execute]
```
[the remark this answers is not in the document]
weissi: I didn't do that because it could be interpreted as `.read` only, not "all the read-only attributes" wdyt?

Sources/NIONFS3/NFSTypes+Common.swift
```diff
@@ -0,0 +1,600 @@
...
+}
+
+public enum NFS3FileType: UInt32 {
```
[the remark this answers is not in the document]
weissi: dunno, `ftype3` lists them like that: https://www.rfc-editor.org/rfc/rfc1813#page-20

Sources/NIONFS3/NFSTypes+Mount.swift
weissi: fantastic catch, thank you! removed the NFS ones, they're MOUNT procedures

Sources/NIONFS3/DummyFS.swift
```diff
@@ -0,0 +1,264 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+final class DummyFS: NFS3FileSystemNoAuth {
```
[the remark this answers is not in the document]
weissi: no, removed for now.

Sources/NIONFS3/MountTypes+Mount.swift
```diff
@@ -0,0 +1,82 @@
...
+    public var result: NFS3Result<Okay, NFS3Nothing>
+}
+
+extension ByteBuffer {
+    public mutating func readNFSCallMount() throws -> MountCallMount {
+        let dirPath = try self.readNFSString()
+        return MountCallMount(dirPath: dirPath)
+    }
+
+    public mutating func writeNFSCallMount(_ call: MountCallMount) {
```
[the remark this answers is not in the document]
weissi: all done now

Sources/NIONFS3/NFSTypes+Common.swift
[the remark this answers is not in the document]
glbrntt: Is there ever room for a custom status? I.e. should this be a struct?

Sources/NIONFS3/NFSTypes+Common.swift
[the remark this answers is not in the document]
glbrntt: Ah so the point being that it won't be extended?

Sources/NIONFS3/NFSFileSystemServerHandler.swift
```diff
@@ -0,0 +1,124 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+
+public final class NFS3FileSystemServerHandler<FS: NFS3FileSystemNoAuth> {
```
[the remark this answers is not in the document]
glbrntt: This one was missed @weissi

Sources/NIONFS3/NFSTypes+Common.swift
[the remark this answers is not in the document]
glbrntt: Feels a bit like reinventing the wheel but fair enough.

Sources/NIONFS3/NFSTypes+Common.swift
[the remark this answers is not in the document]
glbrntt: Having a type which matches the protocol is _definitely_ valuable.

I just don't think we should do it with typealiases. Would much rather we have structs for each of these and actually enforce the type safety. (sorry!)

Sources/NIONFS3/NFSTypes+Common.swift
```diff
@@ -0,0 +1,608 @@
...
+            attributes.ctime.seconds,
+            attributes.ctime.nanoseconds,
+            endianness: .big)
+    }
+
+    @discardableResult public mutating func writeNFSBool(_ bool: NFS3Bool) -> Int {
+        self.writeInteger(bool == true ? 1 : 0, endianness: .big, as: UInt32.self)
+    }
+
+    public mutating func readNFSBool() throws -> Bool {
```
Lukasa, line 502: Should this return `nil` if there are insufficient bytes?

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+        return self.setInteger(header.rawValue, at: index, endianness: .big)
+    }
+
+    @discardableResult public mutating func writeRPCFragmentHeader(_ header: RPCFragmentHeader) -> Int {
+        let bytesWritten = self.setRPCFragmentHeader(header, at: self.writerIndex)
+        self.moveWriterIndex(forwardBy: bytesWritten)
+        return bytesWritten
+    }
+
+    mutating func readRPCReply(xid: UInt32) throws -> RPCReply {
```
Lukasa, line 208: Should this return `nil` if there are insufficient bytes?

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+            call.program,
+            call.programVersion,
+            call.procedure,
+            endianness: .big)
+        + self.writeRPCCredentials(call.credentials)
+        + self.writeRPCVerifier(call.verifier)
+    }
+
+    @discardableResult public mutating func writeRPCReply(_ reply: RPCReply) -> Int {
+        var bytesWritten = self.writeInteger(RPCMessageType.reply.rawValue, endianness: .big)
```
Lukasa, line 273:
```suggestion
        var bytesWritten = self.writeInteger(RPCMessageType.reply.rawValue)
```

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+
+        switch reply.status {
+        case .messageAccepted(_):
+            bytesWritten += self.writeInteger(0 /* accepted */, endianness: .big, as: UInt32.self)
```
Lukasa, line 277:
```suggestion
            bytesWritten += self.writeInteger(0 /* accepted */, as: UInt32.self)
```

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+        case .messageDenied(_):
+            // FIXME: MSG_DENIED (spec name) isn't actually handled correctly here.
+            bytesWritten += self.writeInteger(1 /* denied */, endianness: .big, as: UInt32.self)
```
Lukasa, line 280:
```suggestion
            bytesWritten += self.writeInteger(1 /* denied */, as: UInt32.self)
```

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+        }
+        bytesWritten += self.writeInteger(0 /* verifier */, endianness: .big, as: UInt64.self)
```
Lukasa, line 282:
```suggestion
        bytesWritten += self.writeInteger(0 /* verifier */, as: UInt64.self)
```

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+        + self.writeInteger(0 /* executed successfully */, endianness: .big, as: UInt32.self)
```
Lukasa, line 283:
```suggestion
        + self.writeInteger(0 /* executed successfully */, as: UInt32.self)
```

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+        return bytesWritten
+    }
+
+
+    public mutating func readRPCCall(xid: UInt32) throws -> RPCCall {
+        guard let version = self.readInteger(endianness: .big, as: UInt32.self),
```
Lukasa, line 289:
```suggestion
        guard let version = self.readInteger(as: UInt32.self),
```

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+              let program = self.readInteger(endianness: .big, as: UInt32.self),
```
Lukasa, line 290:
```suggestion
              let program = self.readInteger(as: UInt32.self),
```

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+              let programVersion = self.readInteger(endianness: .big, as: UInt32.self),
```
Lukasa, line 291:
```suggestion
              let programVersion = self.readInteger(as: UInt32.self),
```

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+              let procedure = self.readInteger(endianness: .big, as: UInt32.self) else {
```
Lukasa, line 292:
```suggestion
              let procedure = self.readInteger(as: UInt32.self) else {
```

Sources/NIONFS3/NFSTypes+Containers.swift
Lukasa, line 289: Read multiple integers?

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+        return RPCCall(xid: xid,
+                       rpcVersion: version,
+                       program: program,
+                       programVersion: programVersion,
+                       procedure: procedure,
+                       credentials: credentials,
+                       verifier: verifier)
+    }
+
+    public mutating func readNFSReply(programAndProcedure: RPCNFSProcedureID, rpcReply: RPCReply) throws -> RPCNFS3Reply {
```
Lukasa, line 311: Should this return `nil` if there are insufficient bytes?

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+        case .nfsReadLink:
+            return .init(rpcReply: rpcReply, nfsReply: .readlink(try self.readNFSReplyReadlink()))
+        case .nfsSetAttr:
+            return .init(rpcReply: rpcReply, nfsReply: .setattr(try self.readNFSReplySetattr()))
+        default:
+            throw NFS3Error.unknownProgramOrProcedure(.reply(rpcReply))
+        }
+    }
+
+    mutating func readNFSCall(rpc: RPCCall) throws -> RPCNFS3Call {
```
Lukasa, line 348: Should this return `nil` if there are insufficient bytes?

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+            return bytesWritten
+            &+ self.writeImmutableBuffer(buffer)
+            &+ self.writeRepeatingByte(0x41, count: fillBytes)
+        }
+    }
+
+    public mutating func readRPCMessage() throws -> (RPCMessage, ByteBuffer)? {
+        let save = self
+        guard let fragmentHeader = try self.readRPCFragmentHeader(),
+              let xid = self.readInteger(endianness: .big, as: UInt32.self),
```
Lukasa, line 501:
```suggestion
              let xid = self.readInteger(as: UInt32.self),
```

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+              let messageType = self.readInteger(endianness: .big, as: UInt32.self) else {
```
Lukasa, line 502:
```suggestion
              let messageType = self.readInteger(as: UInt32.self) else {
```

Lukasa (changes requested):

Package.swift
```diff
@@ -108,7 +108,20 @@ var targets: [PackageDescription.Target] = [
             "NIOSOCKS",
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOEmbedded", package: "swift-nio"),
-        ])
+        ]),
+    .target(
```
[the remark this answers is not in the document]
weissi: @Lukasa for now, yes. In a follow-up I might either add a product (either `NIONFS3` or `_NIONFS3Preview`) but we can discuss then. I want to first land the bulk of the code

Sources/NIONFS3/MountTypes+Mount.swift
```diff
@@ -0,0 +1,85 @@
...
+
+        public var fileHandle: NFS3FileHandle
+        public var authFlavors: [RPCAuthFlavor] = [.unix]
+    }
+
+    public var result: NFS3Result<Okay, NFS3Nothing>
+}
+
+extension ByteBuffer {
+    public mutating func readNFSCallMount() throws -> MountCallMount {
```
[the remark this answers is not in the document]
weissi: No because NFS3 is always within the SUN RPC protocol which checks the lengths. So this cannot read short.

Sources/NIONFS3/NFSTypes+Containers.swift
weissi: same as above: no (because it's inside SUN RPC)

Sources/NIONFS3/NFSTypes+Containers.swift
weissi: SUN RPC

Sources/NIONFS3/MountTypes+Mount.swift
```diff
@@ -0,0 +1,85 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+// MARK: - Mount
+public struct MountCallMount: Equatable {
```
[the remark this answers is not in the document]
weissi: Hashables are all done

Sources/NIONFS3/NFSTypes+Containers.swift
weissi: @Lukasa I though explicit over implicit, but can remove. wdyt?

Sources/NIONFS3/NFSFileSystemHandler.swift
```diff
@@ -0,0 +1,96 @@
...
+            let reply = RPCNFS3Reply(rpcReply: .init(xid: call.rpcCall.xid,
+                                                     status: self.rpcReplySuccess),
+                                     nfsReply: reply)
+            context.writeAndFlush(self.wrapOutboundOut(reply), promise: nil)
+        }
+    }
+
+    func sendError(_ error: Error, call: RPCNFS3Call) {
+        if let context = self.context {
+            context.fireErrorCaught(error)
```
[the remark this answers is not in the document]
weissi: good call, thanks, done

Sources/NIONFS3/NFSTypes+Containers.swift
Lukasa: I honestly read it as just noise. Big-endian is overwhelmingly the default, so I don't tend to need anyone to tell me that. I _do_ need them to tell me when they're doing something unusual though.

Sources/NIONFS3/NFSTypes+Common.swift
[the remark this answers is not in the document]
weissi: @Lukasa / @glbrntt done.

Sources/NIONFS3/NFSTypes+Common.swift
```diff
@@ -0,0 +1,608 @@
...
+    case blockDevice = 3
+    case characterDevice = 4
+    case link = 5
+    case socket = 6
+    case fifo = 7
+}
+
+public typealias NFS3Bool = Bool
+
+public struct NFS3FileHandle: Hashable, CustomStringConvertible {
```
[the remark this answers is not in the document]
weissi: done

Sources/NIONFS3/NFSTypes+Containers.swift
```diff
@@ -0,0 +1,529 @@
...
+
+    case _PLEASE_DO_NOT_EXHAUSTIVELY_MATCH_THIS_ENUM_NEW_CASES_MIGHT_BE_ADDED_IN_THE_FUTURE
+}
+
+internal func nfsStringFillBytes(_ byteCount: Int) -> Int {
+    return (4 - (byteCount % 4)) % 4
+}
+
+extension ByteBuffer {
+    mutating func readRPCVerifier() throws -> RPCOpaqueAuth {
```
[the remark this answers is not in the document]
weissi: yes, thanks

Sources/NIONFS3/NFSTypes+Containers.swift
weissi: ok

weissi: done

Package.swift
dnadoba: @weissi FYI: Swift 5.8 has a new flag to tell you if you forget a dependency `--explicit-target-dependency-import-check error`. We have enabled it for the nightly CI for some time now and [we added Swift 5.8. CI today with it enabled. ](https://github.com/apple/swift-nio-extras/blob/e4efedbf47554258e85e6c3949fc956fa6187c76/docker/docker-compose.2204.58.yaml#L14) IIRC the plan is to enable this in Swift 6 by default.

weissi: Sure, I'm aware of that but I don't think it's reasonable to force users to list NIOCore, NIOPosix, NIOEmbedded, ... etc when previously just listing `NIO` was correct?

dnadoba: I strongly disagree with respect to what `import NIO` means today. The set of reexported libraries (NIOCore, NIOPosix, NIOEmbedded) only have historic reasons because they once were all in the original `NIO` module and were not chosen on purpose. Therefore it is only very rarely what you actually want as `NIOEmbedded` is usually only used in tests and not in the actually implementation. The current PR is actually a good example for this. The `NIO` dependency of `NIONFS3` was not replaced with `NIOCore`, `NIOPosix` and `NIOEmbedded` but just with `NIOCore`. The test target has replaced `NIO` with `NIOCore` and `NIOEmbedded`.

Umbrella modules are not really a common thing in Swift. The only way to find out that it is an umbrella module and which libraries it covers is to either look in the documentation or in the Package.swift and implementation. Tooling isn't great either.

I agree that it is anyone to have lots of imports. I don't have a good answer today though. Maybe we should design one or more umbrella modules which make this easier but I can't come up with a good set which include more than 2 libraries. Maybe `NIOPosix`, `NIOHTTP1`, `NIOWebSocket` etc. should just reexport `NIOCore`.

weissi: Also, what if `NIOCore` turns out to not be small enough? It might be split even further but `NIOCore` has to continue to work, much like `NIO` today.
Maybe the SwiftPM feature which allows you to export a product that then exports multiple modules solves part or this issue.

This definitely needs exploration on how we can square usability and technical needs (like not necessarily bringing in NIOPosix)

weissi: Just for the record: happy to change in this PR but I think this is the wrong direction and just messes with users to now have to decide if they want NIO, NIOCore, NIOPosix or maybe something else. They have no idea what NIOPosix or NIOCore even is but they do know `import NIO` because it’s everywhere :)

Lukasa: The problem with this is simply one of compatibility. If you bring NIOPosix with you then you force everyone to compile it, a source of substantial pain for users who don't need or want it. It bloats binary size, increases compile times, and generally makes people's lives harder. We shouldn't make people pay for what they aren't using.

The NIO umbrella module exists for end-users who want the umbrella. For people who are aware of the differences, we should enable the advanced use-case.

weissi: As I said, happy to changed _here_ because it's a NIO package and it's fine to mess with NIO devs (because there are legitimate special case users that don't want posix). But in general, the `NIO` umbrella should be fully supported and the default way to pull in NIO IMHO.

Lukasa (approved):

PR #157, opened 2022-04-21
NIORequestResponseWithIDHandler

weissi:
### Motivation:

`NIORequestResponseHandler` is great, for protocols where responses are in the same order as requests. Unfortunately, it doesn't work for protocols where responses may arrive out-of-order with regards to the order of the requests, such as NFS (#155).

### Modifications:

- Add `NIORequestResponseWithIDHandler`

### Result:

Ability to easily handle protocols with out-of-order responses.


Sources/NIOExtras/TaggedRequestResponseHandler.swift
```diff
@@ -0,0 +1,130 @@
...
+        case .operational:
+            self.promiseBuffer[request.requestID] = responsePromise
+            context.write(self.wrapOutboundOut(request), promise: promise)
+        }
+    }
+}
+
+extension NIOExtrasErrors {
+    public struct ResponseForInvalidRequest<Response: NIORequestIdentifiable>: NIOExtrasError, Equatable {
+        var id: Response.RequestID
```
glbrntt, line 127: This should be `public` and we'll also need an `init`

glbrntt (commented):
Looks good but can we have a few tests?

Sources/NIOExtras/TaggedRequestResponseHandler.swift
weissi: thanks, done!

weissi:
@glbrntt thank you, added the tests. Added all the same ones as for the regular request/response handler plus a few extra tests that test out-of-order and receiving a response that doesn't match (wrong req id).

Sources/NIOExtras/NIORequestIdentifiable.swift
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public protocol NIORequestIdentifiable {
```
glbrntt, line 15: It's a bit sad that `Identifiable` has platform requirements and we can't use that instead.

glbrntt (approved):
Fab, you just need to regenerate the tests @weissi.

Sources/NIOExtras/NIORequestIdentifiable.swift
glbrntt: Seems like I was wrong about this: is there any reason we can't use `Identifiable` instead of this?

weissi: @glbrntt that's what I had before but @Lukasa asked me to change it to this :)

dnadoba: @Lukasa what's the reason not to use `Identifiable`? 
Otherwise I'm not sure if the naming is quite right. This protocol isn't just used to identify requests but responses as well. Maybe we should name it `NIORequestReponseIdentifiable` and `var requestResponseID: RequestResponseID`? But that's quite a mouthful.

glbrntt: To avoid squatting on `var id`, I believe.

dnadoba: I don't really see why this is a problem as one can always create a wrapper type. But I'm fine with it if we rename it to `NIORequestResponseIdentifiable`. 
If we really don't want to decide on a name for the property we could also just let the user give the channel handler two closures in the form of `(Request) -> ID` and `(Response) -> ID` where `ID: Hashable` is an additional generic parameter of the channel handler.

Sources/NIOExtras/TaggedRequestResponseHandler.swift
```diff
@@ -0,0 +1,134 @@
...
+/// `NIORequestIdentifiable`.
+///
+/// `NIOTaggedRequestResponseHandler` does support pipelining `Request`s and it will send them pipelined further down the
+/// `Channel`. Should `RequestResponseHandler` receive an error from the `Channel`, it will fail all promises meant for
+/// the outstanding `Reponse`s and close the `Channel`. All requests enqueued after an error occured will be immediately
+/// failed with the first error the channel received.
+///
+/// `NIOTaggedRequestResponseHandler` does _not_ require that the `Response`s arrive on `Channel` in the same order as
+/// the `Request`s were submitted. They are matched by their `requestID` property (from `NIORequestIdentifiable`).
+public final class NIOTaggedRequestResponseHandler<Request: NIORequestIdentifiable,
```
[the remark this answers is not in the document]
weissi: hmm "Identifiable" sounds weird. Maybe `NIORequestResponseWithIDsHandler`?

Sources/NIOExtras/NIORequestIdentifiable.swift
Lukasa: Yeah, I was reluctant for us to use `var id`, which I tend to think is over-broad. That said, I don't feel this excessively strongly, so I'm open to us just using `Identifiable` instead.

Sources/NIOExtras/TaggedRequestResponseHandler.swift
```diff
@@ -0,0 +1,134 @@
...
+        self.promiseBuffer.reserveCapacity(initialBufferCapacity)
+    }
+
+    public func channelInactive(context: ChannelHandlerContext) {
+        switch self.state {
+        case .error:
+            // We failed any outstanding promises when we entered the error state and will fail any
+            // new promises in write.
+            assert(self.promiseBuffer.count == 0)
+        case .operational:
```
Lukasa, line 72: Do we need a state change here? I suspect the state machine is insufficiently large, because if a user _does_ `write` after we hit `channelInactive` then we'll buffer their promise and then leak it.

Lukasa (changes requested):

Sources/NIOExtras/TaggedRequestResponseHandler.swift
weissi: Yes, thanks, fixed & tested

Sources/NIOExtras/NIORequestIdentifiable.swift
weissi: Decided offline that we're sticking with it as is.

Lukasa (approved):

PR #158, opened 2022-04-21
more PCAP

weissi:
### Motivation:

PCAP is an important protocol. Previously the implementation was very incomplete and split between `NIOWritePCAPHandler` and its tests.

### Modifications:

- Put the implementation into a (private) `NIOPCAP` module.

### Result:

- More PCAP capabilities.


Sources/NIOPCAP/PCAP.swift
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    public init(payloadLength: Int, addresses: AddressTuple, tcp: TCPHeader) {
+        var tv = timeval()
+        gettimeofday(&tv, nil)
+        self = .init(payloadLength: payloadLength, addresses: addresses, time: tv, tcp: tcp)
+    }
+}
+
+public struct PCAPRecord {
```
Lukasa, line 169: Do these three structs need public memberwise initializers?

Sources/NIOPCAP/PCAP.swift
```diff
@@ -0,0 +1,469 @@
...
+    public var dst: in6_addr
+    public var payloadLength: Int
+    public var tcpHeader: TCPHeader
+    public var tcpPayload: ByteBuffer
+}
+
+extension ByteBuffer {
+    // read & parse a TCP packet, containing everything belonging to it (including payload)
+    public mutating func readTCPHeader() throws -> TCPHeader? {
+        struct ParsingError: Error {
```
Lukasa, line 196: Should this error be declared at a higher scope?

Lukasa (approved):
Broadly fine as an internal interface.

Sources/NIOPCAP/PCAP.swift
weissi: thanks, done

Lukasa:
Can you re-run the soundness script and commit the results please?

PR #160, opened 2022-04-28
Implement Sendable

Davidde94:
Make every non-channel `class` conform to `Sendable` (or `@unchecked Sendable`) to prepare for NIO.


Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -13,17 +13,19 @@
...
 import NIOCore
+import NIOConcurrencyHelpers
 
 // MARK: NIOPCAPRingBuffer
 /// Storage for the most recent set of packets captured subject to constraints.
 /// Use `addFragment` as the sink to a `NIOWritePCAPHandler` and call `emitPCAP`
 /// when you wish to get the recorded data.
 /// - Warning:  This class is not thread safe so should only be called from one thread.
-public class NIOPCAPRingBuffer {
+public class NIOPCAPRingBuffer: @unchecked Sendable {
```
glbrntt: Does this class actually need to be `Sendable`? The warning above indicates it isn't thread safe and should be used appropriately (i.e. the caller is responsible for ensuring thread safety)

Davidde94: I guess now it _is_ threadsafe we can remove the warning?

glbrntt: Right, my point is that being sendable isn't always desirable. 

I just want to make sure we're adding it for the right reason (because we think it _should_ be sendable rather than for the sake of making it sendable) given it was intentionally not made thread-safe previously.

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import NIOCore
+import NIOConcurrencyHelpers
```
glbrntt, line 16: Can you add this to the appropriate target in `Package.swift` as well?

Sources/NIOExtras/QuiescingHelper.swift
```diff
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import NIOCore
+import NIOConcurrencyHelpers
```
glbrntt, line 16: No longer used here

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -20,6 +20,7 @@ import Glibc
 import Dispatch
 
 import NIOCore
+import NIOConcurrencyHelpers
```
glbrntt, line 23: No longer used here

Sources/NIOExtrasPerformanceTester/PCAPPerformanceTest.swift
```diff
@@ -16,20 +16,19 @@ import NIOCore
 import NIOEmbedded
 import NIOExtras
 import Foundation
+import NIOConcurrencyHelpers
```
glbrntt, line 19: Not used here

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -24,6 +25,7 @@ public class NIOPCAPRingBuffer {
     private var pcapCurrentBytes: Int
     private let maximumFragments: Int
     private let maximumBytes: Int
+    private let lock = Lock()
```
Lukasa, line 28: Yeah I think this change isn't necessary: this class was fine as it was, it doesn't need to be sendable or thread-safe.

Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
```diff
@@ -15,21 +15,27 @@
 import NIOCore
 import NIOExtras
 import Foundation
+import NIOConcurrencyHelpers
 
 class HTTP1ThreadedPCapPerformanceTest: HTTP1ThreadedPerformanceTest {
-    private class SinkHolder {
+    class SinkHolder {
```
Lukasa, line 21: Did we need to make this non-`private`?

Sources/NIOExtrasPerformanceTester/HTTP1PerformanceTestFramework.swift
```diff
@@ -121,13 +122,17 @@ final class RepeatedRequests: ChannelInboundHandler {
 
 // MARK: ThreadedPerfTest
 class HTTP1ThreadedPerformanceTest: Benchmark {
```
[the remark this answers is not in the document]
Lukasa: Frankly I think this one, along with the benchmarks that inherit from it, can probably just not be `Sendable` at all.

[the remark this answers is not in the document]
Davidde94: Sure, I'll remove :)

Sources/NIOExtras/PCAPRingBuffer.swift
Davidde94: Sorted :)

Sources/NIOExtrasPerformanceTester/HTTP1PerformanceTestFramework.swift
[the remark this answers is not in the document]
Lukasa: We can remove the associated changes too.

Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
Lukasa: Bump

Sources/NIOExtrasPerformanceTester/PCAPPerformanceTest.swift
Lukasa: Bump

Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
Davidde94: @Lukasa nope, this was to be able to extend it (could also have used `fileprivate` I guess, but I'll revert as we're no longer extending.

Sources/NIOExtrasPerformanceTester/HTTP1PerformanceTestFramework.swift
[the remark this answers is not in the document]
Davidde94: I think I've got everything

Davidde94:
@glbrntt @Lukasa I think I've dealt with everything

Sources/NIOExtrasPerformanceTester/HTTP1PCAPPerformanceTests.swift
```diff
@@ -15,6 +15,7 @@
 import NIOCore
 import NIOExtras
 import Foundation
+import NIOConcurrencyHelpers
```
Lukasa, line 18: Unused dependency.

Sources/NIOExtrasPerformanceTester/HTTP1PerformanceTestFramework.swift
```diff
@@ -15,6 +15,7 @@
 import NIOCore
 import NIOPosix
 import NIOHTTP1
+import NIOConcurrencyHelpers
```
Lukasa, line 18: Unused dependency.

Package.swift
```diff
@@ -20,6 +20,7 @@ var targets: [PackageDescription.Target] = [
         name: "NIOExtras",
         dependencies: [
             .product(name: "NIO", package: "swift-nio"),
+            .product(name: "NIOConcurrencyHelpers", package: "swift-nio"),
```
Lukasa, line 23: I don't think we actually use this anymore.

glbrntt (approved):

PR #169, opened 2022-08-02
Docnioextras

PeterAdams-A:
Update docc documentation

### Motivation:

Documentation is good for users.

### Modifications:

Updated docc comments to be more complete.  Linked types where appropriate.

### Result:

Improved docc documentation.


Lukasa (approved):

Lukasa:
Do you mean versions of swift-nio? I don't think that's necessary: there are no relevant correctness issues with old copies of NIO, and we don't _require_ the new ones, so there's no particular issue with letting users resolve older versions if that's necessary to get a buildable tree.

PR #174, opened 2022-08-23
Adopt `Sendable` in `NIOExtras`

dnadoba:
Incremental `Sendable` adoption.

### DIFF
--- a/Sources/NIOExtras/FixedLengthFrameDecoder.swift
+++ b/Sources/NIOExtras/FixedLengthFrameDecoder.swift
@@ -28,14 +28,17 @@ import NIOCore
 ///     | ABC | DEF | GHI |
 ///     +-----+-----+-----+
 ///
-public final class FixedLengthFrameDecoder: ByteToMessageDecoder {
+public final class FixedLengthFrameDecoder: ByteToMessageDecoder, NIOSendable {
     /// Data type we receive.
     public typealias InboundIn = ByteBuffer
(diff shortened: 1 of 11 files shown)


Sources/NIOExtras/FixedLengthFrameDecoder.swift
```diff
@@ -28,14 +28,17 @@ import NIOCore
...
     public typealias InboundIn = ByteBuffer
     /// Data type we send to the next stage.
     public typealias InboundOut = ByteBuffer
 
     @available(*, deprecated, message: "No longer used")
-    public var cumulationBuffer: ByteBuffer?
+    public var cumulationBuffer: ByteBuffer? {
+        get { nil }
+        set { /* discard newValue */ }
+    }
```
Lukasa, line 41: I don't think we can safely make this behavioural change. Given that there's no particular reason to require this type to be `Sendable`, I propose we mark it "explicitly not `Sendable`" and tolerate that.

Lukasa (changes requested):

Sources/NIOExtras/FixedLengthFrameDecoder.swift
dnadoba: Agree, that's the better solution. Shortly thought about making `FixedLengthFrameDecoder` a struct instead of a class but that would be a source breaking change too.

Changed in daae52c

Lukasa (approved):

Issue #176, opened 2022-09-14
Deflate decompression hangs on trailing garbage


vojtarylko:
I tried to extend `HTTPRequestDecompressorTest` with this test case:

```swift
    func testDecompressionTrailingData() throws {
        // Valid compressed data with some trailing garbage
        let compressed = ByteBuffer(bytes: [120, 156, 99, 0, 0, 0, 1, 0, 1] + [1, 2, 3])

        let channel = EmbeddedChannel()
        try channel.pipeline.addHandler(NIOHTTPRequestDecompressor(limit: .none)).wait()
        let headers = HTTPHeaders([("Content-Encoding", "deflate"), ("Content-Length", "\(compressed.readableBytes)")])
        try channel.writeInbound(HTTPServerRequestPart.head(.init(version: .init(major: 1, minor: 1), method: .POST, uri: "https://nio.swift.org/test", headers: headers)))

        try channel.writeInbound(HTTPServerRequestPart.body(compressed))
    }
```

and when tried to run it, the test hangs forever.

If confirmed, this is security issue which can cause DoS.

Lukasa:
Great catch, resolved by #177. I'll try to get a fix out ASAP. I'll also retroactively apply for a CVE and publish a notification, but as this was publicly reported I'd like to get the release out first. Thanks so much for reporting!

Lukasa:
Fix is out.

vojtarylko:
@Lukasa Thanks for fast fix! Was CVE published already?

Lukasa:
Not yet, I’ll update this issue when it is.

Lukasa:
We've assigned this CVE-2022-3252.

PR #177, opened 2022-09-15
Correctly validate the bounds of decompression

### LINKED ISSUE #175: Deflate decompression doesn't fail for truncated input
I tried to extend `HTTPRequestDecompressorTest` with this test case:

```swift
    func testDecompressionTruncatedInput() throws {
        // Truncated compressed data
        let compressed = ByteBuffer(bytes: [120, 156, 99, 0])

        try channel.pipeline.addHandler(NIOHTTPRequestDecompressor(limit: .none)).wait()
        let headers = HTTPHeaders([("Content-Encoding", "deflate"), ("Content-Length", "\(compressed.readableBytes)")])
        try channel.writeInbound(HTTPServerRequestPart.head(.init(version: .init(major: 1, minor: 1), method: .POST, uri: "https://nio.swift.org/test", headers: headers)))

        do {
            try channel.writeInbound(HTTPServerRequestPart.body(compressed))
            XCTFail("writeInbound should fail")
        } catch let error as NIOHTTPDecompression.DecompressionError {
            switch error {
            case .inflationError(Int(Z_BUF_ERROR)):
                // ok
                break
            default:
                XCTFail("Unexptected error: \(error)")
            }
        }
    }
```

As you can see I'm sending *truncated* invalid compressed data. I'd expect to get `DecompressionError.inflationError(-5)`. Instead decompression succeeds.

(I guess root cause is in `z_stream_s.inflatePart(to:minimumCapacity:)` which always calls `inflate` with `Z_NO_FLUSH` parameter.)

Is the problem in my test case, or in the NIO code?

Lukasa:
Motivation

Currently we don't confirm that the decompression has completed successfully. This means that we can incorrectly spin forever attempting to decompress past the end of a message, and that we can fail to notice that a message is truncated. Neither of these is good.

Modifications

Propagate the message zlib gives us as to whether or not decompression is done, and keep track of it.
Add some tests written by @vojtarylko to validate the behaviour.

Result

Correctly police the bounds of the messages.
Resolves #175 and #176.

### DIFF
--- a/Sources/NIOHTTPCompression/HTTPDecompression.swift
+++ b/Sources/NIOHTTPCompression/HTTPDecompression.swift
@@ -57,6 +57,29 @@ public enum NIOHTTPDecompression {
         case initializationError(Int)
     }
 
+    public struct ExtraDecompressionError: Error, Hashable, CustomStringConvertible {
+        private var backing: Backing
+
+        private enum Backing {
+            case invalidTrailingData
+            case truncatedData
+        }
+
+        private init(_ backing: Backing) {
+            self.backing = backing
+        }
+
+        /// Decompression completed but there was invalid trailing data behind the compressed data.
+        public static let invalidTrailingData = Self(.invalidTrailingData)
+
+        /// The decompressed data was incorrectly truncated.
+        public static let truncatedData = Self(.truncatedData)
+
+        public var description: String {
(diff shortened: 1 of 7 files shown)


FranzBusch (approved):
LGTM!

PR #184, opened 2022-10-28
Use #fileID/#filePath instead of #file

carolinacass:
Use #fileID/#filePath instead of #file

Motivation:

#fileID introduced in Swift 5.3, so no longer need to use #file anywhere                                  

Modifications:

Changed #file to #filePath or #fileID depending on situation


glbrntt (approved):
Thank you!

glbrntt:

PR #185, opened 2022-11-30
Migrate http1 proxy connect handler

rnro:
### Motivation:

Moving the `HTTP1ProxyConnectHandler` into `swift-nio-extras` will make the
code which is generally useful when dealing with HTTP1 proxies available
more easily to a wider audience.

### Modifications:

The code and tests are copied over from https://github.com/swift-server/async-http-client/blob/0b5bec741bfcf941e208d937de2ec29affe750a7/Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift.

The code is slightly modified to allow the specification of arbitrary headers rather than authorisation information.

### Result:

`HTTP1ProxyConnectHandler` will be surfaced via the `NIOExtras` library


Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,226 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+import NIOHTTP1
+
+public final class HTTP1ProxyConnectHandler: ChannelDuplexHandler, RemovableChannelHandler {
```
glbrntt, line 18: We need to prefix this with `NIO`: https://github.com/apple/swift-nio/blob/main/docs/public-api.md#1-no-global-namespace-additions-that-arent-prefixed

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,226 @@
...
+    public typealias OutboundIn = Never
+    public typealias OutboundOut = HTTPClientRequestPart
+    public typealias InboundIn = HTTPClientResponsePart
+
+    enum State {
```
glbrntt, line 23: `private`?

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,226 @@
...
+
+    private var state: State = .initialized
+
+    private let targetHost: String
+    private let targetPort: Int
+    private let headers: HTTPHeaders
+    private let deadline: NIODeadline
+
+    private var proxyEstablishedPromise: EventLoopPromise<Void>?
+    var proxyEstablishedFuture: EventLoopFuture<Void>? {
```
glbrntt, line 44: Is this just for testing or is it meant to be public API?

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,226 @@
...
+        return self.proxyEstablishedPromise?.futureResult
+    }
+
+    init(targetHost: String,
```
glbrntt, line 48: This needs to be `public` too

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,226 @@
...
+            break
+        case .initialized, .connectSent, .headReceived:
+            self.state = .failed(Error.noResult)
+            self.proxyEstablishedPromise?.fail(Error.noResult)
+        }
+    }
+
+    public func channelActive(context: ChannelHandlerContext) {
+        self.sendConnect(context: context)
+    }
```
glbrntt, line 76: I think we should be calling `context.fireChannelActive()` here too.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,226 @@
...
+        case .initialized:
+            preconditionFailure("How can we receive a channelInactive before a channelActive?")
+        case .connectSent(let timeout), .headReceived(let timeout):
+            timeout.cancel()
+            self.failWithError(Error.remoteConnectionClosed, context: context, closeConnection: false)
+
+        case .failed, .completed:
+            break
+        }
+    }
```
glbrntt, line 89: We should be calling `context.fireChannelInactive()` here.

glbrntt (changes requested):
Couple of nits but looks good otherwise

glbrntt:

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
dnadoba: I think we should let the user pass in a promise instead. This also allows `proxyEstablishedFuture` to be non-nil as we would no longer need to delay creation of the promise until the `ChannelHandler` is added to the pipeline to get hold of the `EventLoop`.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
rnro: I think these type aliases have to be public because `(NIO)HTTP1ProxyConnectHandler` is public

glbrntt: The type aliases certainly have to be public, the `private` comment was for `State`

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
glbrntt: Is there any reason we shouldn't use a user inbound event instead if it's not just for testing? (I'm okay keeping a promise/future as well.)

dnadoba: It is not just for testing:
https://github.com/swift-server/async-http-client/blob/5bee16a79922e3efcb5cea06ecd27e6f8048b56b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool%2BFactory.swift#L211-L221

rnro: I added an `EventLoopPromise` to the initialiser.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,232 @@
...
+         promise: EventLoopPromise<Void>?) {
+        self.targetHost = targetHost
+        self.targetPort = targetPort
+        self.headers = headers
+        self.deadline = deadline
+        self.promise = promise
+    }
+
+    public func handlerAdded(context: ChannelHandlerContext) {
+        self.sendConnect(context: context)
```
Lukasa, line 65: This should only happen if `context.channel.isActive`.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,232 @@
...
+        )
+
+        context.write(self.wrapOutboundOut(.head(head)), promise: nil)
+        context.write(self.wrapOutboundOut(.end(nil)), promise: nil)
+        context.flush()
+    }
+
+    private func handleHTTPHeadReceived(_ head: HTTPResponseHead, context: ChannelHandlerContext) {
+        guard case .connectSent(let scheduled) = self.state else {
+            preconditionFailure("HTTPDecoder should throw an error, if we have not send a request")
```
Lukasa, line 147: This `preconditionFailure` is really aggressive. We should absolutely tolerate receiving the HTTP head _late_.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,232 @@
...
+            // inbound proxies) will switch to tunnel mode immediately after the
+            // blank line that concludes the successful response's header section
+            self.state = .headReceived(scheduled)
+        case 407:
+            self.failWithError(Error.proxyAuthenticationRequired, context: context)
+
+        default:
+            // Any response other than a successful response indicates that the tunnel
+            // has not yet been formed and that the connection remains governed by HTTP.
+            self.failWithError(Error.invalidProxyResponse, context: context)
```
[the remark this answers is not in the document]
Lukasa: Yeah, I think this would be a good idea.

Lukasa (changes requested):
Can we move the changed `generate_linux_tests.rb` script and its associated changes to a new PR? They add _loads_ of noise to this one, and they are unrelated.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
rnro: Changed so that we now tolerate the state being either `.failed` or `.completed` at this point without failing a precondition.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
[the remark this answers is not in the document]
rnro: Good idea, I have added this.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
dnadoba: I think we shouldn't tolerate `.completed` either, just `.failed`. We can only be in completed after we have received a http head and shouldn't receive a second head.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,242 @@
...
+            context.close(mode: .all, promise: nil)
+        }
+    }
+
+    /// Error types for ``HTTP1ProxyConnectHandler``
+    public struct Error: Swift.Error, Equatable {
+
+        fileprivate enum ErrorEnum: Equatable {
+            case proxyAuthenticationRequired
+            case invalidProxyResponseHead(head: HTTPResponseHead)
```
dnadoba, line 214: I think we may want to add a custom `Equatable` implementation and ignore the `HTTPResponseHead` in the comparison. Therefore two `invalidProxyResponseHead` errors are equal, regardless of the exact `head`. We have already done something similar if we include the file and line number where we ignore them in the `Equatable` and `Hashable` implementation. @Lukasa what do you think about that?

In addition, `Hashable` conformance would be nice as well.

Lukasa: Yes, that seems wise.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,299 @@
...
+            init(file: String, line: UInt) {
+                self.file = file
+                self.line = line
+            }
+        }
+
+        fileprivate let error: Storage
+        let location: Location
+
+        fileprivate init(error: Storage, file: String = #file, line: UInt = #line) {
```
glbrntt, line 269: Unfortunately we need all the `static let`s to be `static func`s below which also take a `file` and `line` so that we capture the `file` and `line` where the error was created (otherwise it will just be the line of the `static let`).

I recommend removing the defaults here: to enforce that we pass through through to the init.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,299 @@
...
+                self.failWithError(Error.proxyAuthenticationRequired, context: context)
+
+            default:
+                // Any response other than a successful response indicates that the tunnel
+                // has not yet been formed and that the connection remains governed by HTTP.
+                self.failWithError(Error.invalidProxyResponseHead(head), context: context)
+            }
+        case .failed:
+            break
+        case .initialized, .headReceived, .completed:
```
glbrntt, line 166: @Lukasa how much does the `HTTPDecoder` validate state for us? Presumably a remote peer could arrange for us to hit `.completed` state quite easily. Possibly the `.headReceived` state too depending on what the decoder does.

Lukasa: Quite a lot, it won't emit these parts in the wrong order.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,299 @@
...
+            indirect case invalidProxyResponseHead(head: HTTPResponseHead)
+            case invalidProxyResponse
+            case remoteConnectionClosed
+            case httpProxyHandshakeTimeout
+            case noResult
+
+            // compare only the kind of error, not the associated response head
+            @inlinable
+            static func == (lhs: Self, rhs: Self) -> Bool {
+                Kind(lhs) == Kind(rhs)
```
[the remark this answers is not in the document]
dnadoba: How would you implement `hash(into:)` without `Kind`?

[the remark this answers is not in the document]
Lukasa:
```swift
mutating func hash(into hasher: inout Hasher) {
    switch self {
    case .proxyAuthenticationRequired:
        hasher.combine(0)
    case .invalidProxyResponseHead:
        hasher.combine(1)
    case .invalidProxyResponse:
        hasher.combine(2)
    case .remoteConnectionClosed:
        hasher.combine(3)
    case .httpProxyHandshakeTimeout:
        hasher.combine(4)
    case .noResult:
        hasher.combine(5)
    }
}
```

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,299 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+import NIOHTTP1
+
+public final class NIOHTTP1ProxyConnectHandler: ChannelDuplexHandler, RemovableChannelHandler {
+    public typealias OutboundIn = Never
```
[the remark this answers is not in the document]
rnro: In the latest commit I have changed the design to buffer writes until the CONNECT req/res has completed.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,299 @@
...
+            preconditionFailure("Invalid state: \(self.state)")
+        }
+    }
+
+    private func handleHTTPEndReceived(context: ChannelHandlerContext) {
+        switch self.state {
+        case .headReceived(let timeout):
+            timeout.cancel()
+            self.state = .completed
+            self.promise.succeed(())
```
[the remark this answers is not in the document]
rnro: Added in the latest commit

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,299 @@
...
+        case completed
+    }
+
+    private var state: State = .initialized
+
+    private let targetHost: String
+    private let targetPort: Int
+    private let headers: HTTPHeaders
+    private let deadline: NIODeadline
+    private let promise: EventLoopPromise<Void>
```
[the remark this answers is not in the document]
rnro: I have made the promise optional again in the latest commit.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,299 @@
...
+        }
+    }
+
+    public func handlerRemoved(context: ChannelHandlerContext) {
+        switch self.state {
+        case .failed, .completed:
+            break
+        case .initialized, .connectSent, .headReceived:
+            self.state = .failed(Error.noResult)
+            self.promise.fail(Error.noResult)
```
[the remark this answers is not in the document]
rnro: This code was inherited from the original implementation. After checking through the code I believe it is safe to consolidate this logic.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,232 @@
...
+        case completed
+    }
+
+    private var state: State = .initialized
+
+    private let targetHost: String
+    private let targetPort: Int
+    private let headers: HTTPHeaders
+    private let deadline: NIODeadline
+    private let promise: EventLoopPromise<Void>?
```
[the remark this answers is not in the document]
Lukasa: Not sure this got addressed.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,299 @@
...
+
+    public func channelActive(context: ChannelHandlerContext) {
+        self.sendConnect(context: context)
+        context.fireChannelActive()
+    }
+
+    public func channelInactive(context: ChannelHandlerContext) {
+        switch self.state {
+        case .initialized:
+            preconditionFailure("How can we receive a channelInactive before a channelActive?")
```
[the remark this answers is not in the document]
Lukasa: Same here.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,399 @@
...
+            }
+        }
+    }
+
+}
+
+extension NIOHTTP1ProxyConnectHandler.Error: Hashable {
+    // compare only the kind of error, not the associated response head
+    public static func == (lhs: Self, rhs: Self) -> Bool {
+        switch (lhs.store.details, rhs.store.details) {
```
glbrntt, line 352: You could just compare `errorCode` here to simplify this...

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,399 @@
...
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(self.errorCode)
+    }
+}
+
+
+extension NIOHTTP1ProxyConnectHandler.Error: CustomStringConvertible {
+    public var description: String {
+        self.store.details.description
```
glbrntt, line 378: I think we should include the file and line in the error description.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,399 @@
...
+    }
+}
+
+extension NIOHTTP1ProxyConnectHandler.Error.Details: CustomStringConvertible {
+    public var description: String {
+        switch self {
+        case .proxyAuthenticationRequired:
+            return "Proxy Authentication Required"
+        case .invalidProxyResponseHead:
+            return "Invalid Proxy Response Head"
```
dnadoba, line 388: you want to include the `HTTPResponseHead` in the description e.g.: 
```suggestion
        case .invalidProxyResponseHead(let head):
            return "Invalid Proxy Response Head \(head)"
```

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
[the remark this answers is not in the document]
glbrntt: This hasn't been addressed either

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,384 @@
...
+            // write remainder
+            while let bufferedPart = self.bufferedWrittenMessages.popFirst() {
+                context.write(bufferedPart.data, promise: bufferedPart.promise)
+            }
+        }
+
+        context.leavePipeline(removalToken: removalToken)
+    }
+
+
```
glbrntt, line 116:
```suggestion

```

glbrntt (commented):
I think I'm more or less happy with this once https://github.com/apple/swift-nio-extras/pull/185/files#r1039437675 has been addressed.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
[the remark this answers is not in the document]
rnro: Hopefully it is now

dnadoba (approved):
LGTM

glbrntt (approved):
Looks good, thanks @rnro!

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,393 @@
...
+            // flush any messages up to the mark
+            if hadMark {
+                context.flush()
+            }
+
+            // write remainder
+            while let bufferedPart = self.bufferedWrittenMessages.popFirst() {
+                context.write(bufferedPart.data, promise: bufferedPart.promise)
+            }
```
Lukasa, line 110: There's a fairly subtle bug available here where we can inadvertently lose a flush. If one of these calls to `context.write` triggers a flush call, that flush will be lost, and so the writes will hang.

It's probably not worth being this clever, tbh: we should unconditionally flush at the end of the unbuffered writes, even if it's a little inefficient.

Sources/NIOExtras/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,393 @@
...
+        if context.channel.isActive {
+            self.sendConnect(context: context)
+        }
+    }
+
+    public func handlerRemoved(context: ChannelHandlerContext) {
+        switch self.state {
+        case .failed, .completed:
+            // we don't expect there to be any buffered messages in these states
+            assert(self.bufferedWrittenMessages.isEmpty)
```
Lukasa, line 126: This case shouldn't pass silently: we should either throw an error, precondition, or handle the edge-case.

Lukasa (changes requested):

Lukasa (approved):

PR #188, opened 2022-12-02
update generated test script from swift-nio

rnro:
The newer version of the script automatically calculates accurate copyright date statements.


Lukasa:
It looks like the soundness script needs updating to tolerate the new years ranges.

rnro:
> It looks like the soundness script needs updating to tolerate the new years ranges.

Yep, those changes should be there now

Lukasa (approved):

PR #192, opened 2023-02-08
ServerQuiescingHelper no longer leaking promises

carolinacass:
Motivation:
ServerQuiescingHelper leaked promises when promise left scope and not succeeded

Modifications:
Failing promise within a deinit

### DIFF
--- a/Sources/NIOExtras/QuiescingHelper.swift
+++ b/Sources/NIOExtras/QuiescingHelper.swift
@@ -216,6 +216,7 @@ private final class CollectAcceptedChannelsHandler: ChannelInboundHandler {
 ///     try fullyShutdownPromise.futureResult.wait()
 ///
 public final class ServerQuiescingHelper {
+    public struct UnusedQuiescingHelperError: Error {}
     private let channelCollectorPromise: EventLoopPromise<ChannelCollector>
 
     /// Initialize with a given `EventLoopGroup`.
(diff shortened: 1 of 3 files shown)


Sources/NIOExtras/QuiescingHelper.swift
```diff
@@ -226,6 +227,11 @@ public final class ServerQuiescingHelper {
         self.channelCollectorPromise = group.next().makePromise()
     }
 
+    /// Deinitilaize the `channelCollectorPromise` with failure when promise never succeeded
```
glbrntt, line 230: We can remove this comment

Sources/NIOExtras/QuiescingHelper.swift
```diff
@@ -216,6 +216,7 @@ private final class CollectAcceptedChannelsHandler: ChannelInboundHandler {
 ///     try fullyShutdownPromise.futureResult.wait()
 ///
 public final class ServerQuiescingHelper {
+    public struct UnusedQuiescingHelperError: Error {}
```
glbrntt, line 219: We should document this to say something like:

```suggestion
    /// The `ServerQuiescingHelper` was never used to create a channel handler.
    public struct UnusedQuiescingHelperError: Error {}
```

glbrntt (changes requested):
You'll need to run `./scripts/generate_linux_tests.rb` as well

glbrntt (commented):
This one still hasn't been addressed: https://github.com/apple/swift-nio-extras/pull/192#pullrequestreview-1289327247

glbrntt (commented):
You need to run `./scripts/generate_linux_tests.rb` again, for some reason the date didn't update and the soundness check is sad.

glbrntt (approved):

Lukasa (approved):

PR #197, opened 2023-04-12
Drop Swift 5.5

Lukasa:
Motivation

Per SwiftNIO's formal version policy, we are ready to drop support for Swift 5.5.

Modifications

This patch removes the support for 5.5 and all supporting infrastructure. This includes the test generation functionality, which is no longer required, as well as the files generated by that functionality. It updates the dockerfile for 5.8, and it removes all conditional compilation checks that are now definitionally true.

Result

A nice, clean, 5.6+ codebase


Lukasa:
Hey @yim-lee, can we remove the 5.5 job from this repo? It's no longer necessary with this patch.

PR #202, opened 2023-07-05
Add HTTP types adapter for SwiftNIO

guoye-zhang:
Add [swift-http-types](https://github.com/apple/swift-http-types) adapters that works with SwiftNIO HTTP protocols.

### Motivation:

SwiftNIO Channel implementations should be able to use the shared HTTP currency types.

### Modifications:

3 new targets:

* `NIOHTTPTypes`: version independent data types
* `NIOHTTPTypesHTTP1`: convertors between http1 types and new http types
* `NIOHTTPTypesHTTP2`: convertors between http2 types and new http types


tomerd:

tomerd:
@guoye-zhang I think this is missing the dependency?

guoye-zhang:
Oh we need to reduce the swift version of swift-http-types to 5.6

guoye-zhang:
I updated the Package.swift to only depend on swift-http-types on 5.8 or later since it depends on a few new language features.

tomerd:
cc @Lukasa

tomerd (approved):

Package.swift
```diff
@@ -125,16 +137,52 @@ var targets: [PackageDescription.Target] = [
...
+#if swift(>=5.8)
+products += [
+    .library(name: "NIOHTTPTypes", targets: ["NIOHTTPTypes"]),
+    .library(name: "NIOHTTPTypesHTTP1", targets: ["NIOHTTPTypesHTTP1"]),
+    .library(name: "NIOHTTPTypesHTTP2", targets: ["NIOHTTPTypesHTTP2"]),
+]
+
+dependencies += [
+    .package(url: "https://github.com/apple/swift-http-types", from: "0.1.0"),
+]
```
FranzBusch, line 149: We can't merge this until we tag a `1.0.0` in swift-http-types

Sources/NIOHTTPTypes/NIOHTTPTypes.swift
```diff
@@ -0,0 +1,44 @@
...
+import HTTPTypes
+import NIOCore
+
+/// The parts of a complete HTTP message, either request or response.
+///
+/// An HTTP message is made up of a request, or one or more response headers,
+/// encoded by `.head`, zero or more body parts, and optionally some trailers. To
+/// indicate that a complete HTTP message has been sent or received, we use `.end`,
+/// which may also contain any trailers that make up the message.
+public enum HTTPTypePart<HeadT: Equatable, BodyT: Equatable> {
```
FranzBusch, line 24:
```suggestion
public enum HTTPTypePart<Head: Equatable, Body: Equatable> {
```

Sources/NIOHTTPTypes/NIOHTTPTypes.swift
```diff
@@ -0,0 +1,44 @@
...
+    case body(BodyT)
+    case end(HTTPFields?)
+}
+
+extension HTTPTypePart: Sendable where HeadT: Sendable, BodyT: Sendable {}
+
+extension HTTPTypePart: Equatable {}
+
+/// The components of an HTTP request from the view of an HTTP client.
+public typealias HTTPTypeClientRequestPart = HTTPTypePart<HTTPRequest, IOData>
```
FranzBusch, line 35: Can we get rid of `IOData` here the problem is that `IOData` is not `Sendable`

Sources/NIOHTTPTypes/NIOHTTPTypes.swift
```diff
@@ -0,0 +1,44 @@
...
+
+/// The components of an HTTP request from the view of an HTTP server.
+public typealias HTTPTypeServerRequestPart = HTTPTypePart<HTTPRequest, ByteBuffer>
+
+/// The components of an HTTP response from the view of an HTTP client.
+public typealias HTTPTypeClientResponsePart = HTTPTypePart<HTTPResponse, ByteBuffer>
+
+/// The components of an HTTP response from the view of an HTTP server.
+public typealias HTTPTypeServerResponsePart = HTTPTypePart<HTTPResponse, IOData>
```
FranzBusch, line 44: Same here

Sources/NIOHTTPTypesHTTP1/HTTP1ToHTTPCodec.swift
```diff
@@ -0,0 +1,109 @@
...
+    public typealias InboundOut = HTTPTypeClientResponsePart
+
+    public typealias OutboundIn = HTTPTypeClientRequestPart
+    public typealias OutboundOut = HTTPClientRequestPart
+
+    /// Initializes a `HTTP1ToHTTPClientCodec`.
+    public init() {}
+
+    public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        switch unwrapInboundIn(data) {
```
FranzBusch, line 32:
```suggestion
        switch self.unwrapInboundIn(data) {
```

Sources/NIOHTTPTypesHTTP1/HTTP1ToHTTPCodec.swift
```diff
@@ -0,0 +1,109 @@
...
+        case .head(let head):
+            do {
+                try context.fireChannelRead(wrapInboundOut(.head(head.newResponse)))
```
FranzBusch, line 35:
```suggestion
                try context.fireChannelRead(self.wrapInboundOut(.head(head.newResponse)))
```

Sources/NIOHTTPTypesHTTP1/HTTP1ToHTTPCodec.swift
```diff
@@ -0,0 +1,109 @@
...
+            } catch {
+                context.fireErrorCaught(error)
+            }
+        case .body(let body):
+            context.fireChannelRead(wrapInboundOut(.body(body)))
+        case .end(let trailers):
+            context.fireChannelRead(wrapInboundOut(.end(trailers?.newFields(splitCookie: false))))
```
FranzBusch, line 42:
```suggestion
            context.fireChannelRead(self.wrapInboundOut(.body(body)))
        case .end(let trailers):
            context.fireChannelRead(self.wrapInboundOut(.end(trailers?.newFields(splitCookie: false))))
```

FranzBusch: Can you add `self.` in the whole diff please.

Sources/NIOHTTPTypes/NIOHTTPTypes.swift
FranzBusch: Maybe we can get rid off the two different client/server types all together then cc @Lukasa

Package.swift
```diff
@@ -125,16 +137,52 @@ var targets: [PackageDescription.Target] = [
         ]),
 ]
 
+#if swift(>=5.8)
```
FranzBusch, line 140: This is a bit unfortunate. What 5.8 language features are we using that we can't use in 5.7? Is it only the `some` keyword? If so it would be great if we could just write this differently and make it support 5.7

Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
```diff
@@ -0,0 +1,203 @@
...
+
+    func newFields(splitCookie: Bool) -> HTTPFields {
+        var fields = HTTPFields()
+        fields.reserveCapacity(count)
+        for (index, field) in enumerated() {
+            if index == 0, field.name.lowercased() == "host" {
+                continue
+            }
+            if let name = HTTPField.Name(field.name) {
+                if splitCookie, name == .cookie, #available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *) {
```
Lukasa, line 126: As a general style nit we prefer to use `&&` where it is possible to do so.

Sources/NIOHTTPTypes/NIOHTTPTypes.swift
guoye-zhang: Is it OK to synchronously read `FileRegion` into `ByteBuffer` when converting an `HTTPPart` into an `HTTPTypePart`?

guoye-zhang: I'll use a `fatalError` for now

Package.swift
guoye-zhang: 5.8 introduced a few buffer methods (SE-0370) which we are using. I could rewrite it to use memcpy

guoye-zhang: https://github.com/apple/swift-http-types/pull/2

Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
guoye-zhang: This was actually changed to `,` by swiftformat. I changed it back.

Sources/NIOHTTPTypes/NIOHTTPTypes.swift
Lukasa: Nope, unfortunately it isn't: doing that I/O on NIO's event loops isn't acceptable.

Sources/NIOHTTPTypesHTTP1/HTTPToHTTP1Codec.swift
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+        switch self.unwrapOutboundIn(data) {
+        case .head(let response):
+            do {
+                try context.write(self.wrapOutboundOut(.head(response.newResponse)), promise: promise)
+            } catch {
+                context.fireErrorCaught(error)
```
FranzBusch, line 111: Same here

Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
```diff
@@ -0,0 +1,207 @@
...
+        }
+        self.init(
+            version: .http1_1,
+            method: HTTPMethod(newRequest.method),
+            uri: path,
+            headers: headers
+        )
+    }
+
+    func newRequest(secure: Bool, splitCookie: Bool) throws -> HTTPRequest {
```
FranzBusch, line 172: We normally spell this as an `init` on `HTTPRequest` that takes an `HTTPRequestHead` and the two other parameters.

Sources/NIOHTTPTypesHTTP2/HTTPTypeConversion.swift
```diff
@@ -0,0 +1,265 @@
...
+extension HPACKIndexing {
+    init(_ newIndexingStrategy: HTTPField.DynamicTableIndexingStrategy) {
+        switch newIndexingStrategy {
+        case .avoid: self = .nonIndexable
+        case .disallow: self = .neverIndexed
+        default: self = .indexable
+        }
+    }
+
+    var newIndexingStrategy: HTTPField.DynamicTableIndexingStrategy {
```
FranzBusch, line 44: Same here. Normally spelled with an `init` on `HTTPField.DynamicTableIndexingStrategy`

Sources/NIOHTTPTypesHTTP2/HTTPTypeConversion.swift
```diff
@@ -0,0 +1,265 @@
...
+    init(_ newTrailers: HTTPFields) {
+        var headers = HPACKHeaders()
+        headers.reserveCapacity(newTrailers.count)
+        for field in newTrailers {
+            headers.add(newField: field)
+        }
+        self = headers
+    }
+
+    var newRequest: HTTPRequest {
```
FranzBusch, line 101: Same here.

Sources/NIOHTTPTypesHTTP2/HTTPTypeConversion.swift
```diff
@@ -0,0 +1,265 @@
...
+                    field.indexingStrategy = indexable.newIndexingStrategy
+                    request.headerFields.append(field)
+                }
+                i = self.index(after: i)
+            }
+            return request
+        }
+    }
+
+    var newResponse: HTTPResponse {
```
FranzBusch, line 196: and here

Sources/NIOHTTPTypesHTTP2/HTTPTypeConversion.swift
```diff
@@ -0,0 +1,265 @@
...
+                    field.indexingStrategy = indexable.newIndexingStrategy
+                    response.headerFields.append(field)
+                }
+                i = self.index(after: i)
+            }
+            return response
+        }
+    }
+
+    var newTrailers: HTTPFields {
```
FranzBusch, line 248: and here

FranzBusch (commented):
Overall I think this is mostly ready. I left a couple of comments inline.

Sources/NIOHTTPTypesHTTP1/HTTP1ToHTTPCodec.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+        switch self.unwrapOutboundIn(data) {
+        case .head(let request):
+            do {
+                try context.write(self.wrapOutboundOut(.head(HTTPRequestHead(request))), promise: promise)
+            } catch {
+                context.fireErrorCaught(error)
+            }
```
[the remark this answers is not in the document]
guoye-zhang: Do we need both `fireErrorCaught` and failing the promise?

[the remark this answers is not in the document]
guoye-zhang: Oh I see other places doing both

Sources/NIOHTTPTypes/NIOHTTPTypes.swift
```diff
@@ -0,0 +1,34 @@
...
+    case head(Head)
+    case body(ByteBuffer)
+    case end(HTTPFields?)
+}
+
+/// The components of an HTTP request.
+public typealias HTTPTypeRequestPart = HTTPTypePart<HTTPRequest>
+
+/// The components of an HTTP response.
+public typealias HTTPTypeResponsePart = HTTPTypePart<HTTPResponse>
```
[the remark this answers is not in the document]
guoye-zhang: I made them separate enums, but requests can support trailers as far as I know

Package.swift
```diff
@@ -125,16 +137,52 @@ var targets: [PackageDescription.Target] = [
         ]),
 ]
 
+#if swift(>=5.7)
```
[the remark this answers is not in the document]
FranzBusch: Interesting what do you need from the patch release? @Lukasa I think we can bump to the patch release right away. At least there is prior art in doing this AFAIK

Sources/NIOHTTPTypes/NIOHTTPTypes.swift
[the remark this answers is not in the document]
FranzBusch: Yeah I learned this today as well after reading the RFC again. Requests can indeed container trailers.

Package.swift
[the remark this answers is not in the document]
guoye-zhang: I think some String API was added in 5.7.1: https://github.com/apple/swift-http-types/issues/20

Package.swift
```diff
@@ -125,16 +137,52 @@ var targets: [PackageDescription.Target] = [
         ]),
 ]
 
+#if swift(>=5.7.1)
```
FranzBusch, line 140: Just drop this conditional completely. We are going to merge this PR after we drop 5.6 and then we will just make `5.7.1` or bottom version.

Sources/NIOHTTPTypes/NIOHTTPTypes.swift
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/// The parts of a complete HTTP response.
+///
+/// An HTTP response message is made up of one or more response headers encoded
+/// by `.head`, zero or more body parts, and optionally some trailers.
+///
+/// To indicate that a complete HTTP message has been sent or received, we use
+/// `.end`, which may also contain any trailers that make up the message.
+public enum HTTPTypeResponsePart: Sendable, Hashable {
```
FranzBusch, line 38: NIT: I am wondering if we need the `HTTPType` prefix or can just go with `HTTPRequestPart` and `HTTPResponsePart`. Those names haven't been used yet in the NIO ecosystem from what I can tell.

guoye-zhang: Updated

Package.swift
guoye-zhang: Fixed

FranzBusch (approved):
LGTM! @Lukasa mind giving this a look as well

guoye-zhang:
Should I bump the Swift version to 5.7.1 in this PR?

FranzBusch:
> Should I bump the Swift version to 5.7.1 in this PR?

Yes let's do that.

guoye-zhang:
There are still a few Dockerfiles mentioning 5.7, but those should mean latest 5.7.x? Do I need to change them?

Lukasa:
They mean latest 5.7.x. That's typically fine, 5.7 just means "floating" in most cases.

Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
```diff
@@ -0,0 +1,210 @@
...
+        self.init(fields)
+    }
+}
+
+extension HTTPFields {
+    init(_ oldHeaders: HTTPHeaders, splitCookie: Bool) {
+        self.init()
+        self.reserveCapacity(count)
+        for (index, field) in oldHeaders.enumerated() {
+            if index == 0, field.name.lowercased() == "host" {
```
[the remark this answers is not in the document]
guoye-zhang: RFC7230 has this sentence which is no longer present in RFC9112:

>    Since the Host field-value is critical information for handling a
>    request, a user agent SHOULD generate Host as the first header field
>    following the request-line.

I guess we can't assume that it is always first. HTTPTypes does not preserve the order of pseudo header fields, so we will lose the order of Host (which might be fine?)

Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
```diff
@@ -0,0 +1,210 @@
...
+                continue
+            }
+            if let name = HTTPField.Name(field.name) {
+                if splitCookie && name == .cookie, #available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *) {
```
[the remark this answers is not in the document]
guoye-zhang: Cookie splitting is a performance optimization and it isn't done today with HTTP2FramePayloadToHTTP1ClientCodec. URLSession only started doing it in iOS 14. It's not ideal, but it's not worth spending a lot of effort into I think

Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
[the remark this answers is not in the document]
guoye-zhang: Fixed in 6da1b1a

Lukasa (approved):
Nice one, thanks @guoye-zhang!

PR #203, opened 2023-07-10
Import HTTP resumable upload sample code

guoye-zhang:
Support HTTP resumable upload.

### Motivation:

Supporting HTTP resumable upload protocol defined in https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-resumable-upload-05

* Interop version 3: iOS 17.0, macOS 14.0
* Interop version 5: iOS 18.0, macOS 15.0
* Interop version 6: iOS 18.1, macOS 15.1

### Modifications:

2 new public classes, `HTTPResumableUploadHandler` and `HTTPResumableUploadContext`, and a few other supporting objects to manage resumable uploads and translate them into regular uploads.


Lukasa (approved):

Lukasa:
@FranzBusch want to take another quick look at this?

Sources/NIOResumableUpload/HTTPResumableUploadHandler.swift
```diff
@@ -0,0 +1,141 @@
...
+    private var eventLoop: EventLoop!
+
+    /// Create an `HTTPResumableUploadHandler` within a given `HTTPResumableUploadContext`.
+    /// - Parameters:
+    ///   - context: The context for this upload handler.
+    ///              Use the same context across upload handlers, as uploads can't resume across different contexts.
+    ///   - channelConfigurator: A closure for configuring the child HTTP server channel.
+    public init(
+        context: HTTPResumableUploadContext,
+        channelConfigurator: @escaping (Channel) -> Void
```
FranzBusch, line 39: We normally call this the `channelInitializer`. Furthermore we normally let this return an `EventLoopFuture` to give people the chance to initialize their channel asynchronously.

FranzBusch (commented):
Left some smaller comments inline but this is not going to be usable from our new async APIs since we lose type information in the pipeline. Furthermore, I think with our new async APIs we would implement all of this in Swift Concurrency in an HTTP Server and not inside the NIO channel pipeline. I am wondering if it makes sense to add this here or just defer it functionality that our general purpose HTTP server should handle. @Lukasa WDYT?

Lukasa:
I think we should add this here: it's in support of sample code that was released with WWDC, and we should be willing to support that code.

Sources/NIOResumableUpload/HTTPResumableUploadContext.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOConcurrencyHelpers
+import NIOCore
+
+/// `HTTPResumableUploadContext` manages ongoing uploads.
+public final class HTTPResumableUploadContext {
```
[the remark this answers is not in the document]
jrflat: Hey I'll be pitching in for Guoye to address feedback. I think `HTTPResumableUploadContext` is public because it provides a nice way for developers to configure unique or optional behaviors of their resumable uploads. In the future, the protocol draft may contain optional features such as progress reporting, and if we support them, this context could be the place to opt in or out (instead of passing more parameters to the `handler`). Guoye may have other considerations, too--I can ask when he's back from vacation.

Sources/NIOResumableUpload/HTTPResumableUploadChannel.swift
```diff
@@ -0,0 +1,251 @@
...
+        self._isActiveAtomic.load(ordering: .relaxed)
+    }
+
+    var _channelCore: ChannelCore {
+        self
+    }
+
+    let eventLoop: EventLoop
+
+    private var autoRead: Bool
```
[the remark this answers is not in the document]
jrflat: Fixed in [46a7e20](https://github.com/apple/swift-nio-extras/pull/203/commits/46a7e20f24628b3bc2f2ad2f5f4f695e71b2df53)

Sources/NIOResumableUpload/HTTPResumableUploadHandler.swift
guoye-zhang: I checked with @Lukasa before about this. Making it asynchronous is possible but it would definitely complicate the code by adding a pause after completing the header read. This is guaranteed to be invoked no more than once, so they can do the expensive work upfront and capture the result into this closure.

Sources/NIOResumableUpload/HTTPResumableUploadChannel.swift
```diff
@@ -0,0 +1,251 @@
...
+    init(
+        upload: HTTPResumableUpload,
+        parent: Channel,
+        channelConfigurator: (Channel) -> Void
+    ) {
+        self.upload = upload
+        self.allocator = parent.allocator
+        self.closePromise = parent.eventLoop.makePromise()
+        self.eventLoop = parent.eventLoop
+        self.autoRead = try! parent.syncOptions!.getOption(ChannelOptions.autoRead)
```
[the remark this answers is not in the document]
guoye-zhang: I was told that any reasonable Channel should implement `syncOptions` and we can refuse to work with channels that do not. In general, `autoRead` should be true and handlers can defer a read by intercepting `read()` calls. Alternatively, we could just say our `autoRead` defaults to true and not inherit it from the parent channel.

Sources/NIOResumableUpload/HTTPResumableUploadHandler.swift
```diff
@@ -0,0 +1,141 @@
...
+/// A channel handler that translates resumable uploads into regular uploads, and passes through
+/// other HTTP traffic.
+public final class HTTPResumableUploadHandler: ChannelDuplexHandler {
+    public typealias InboundIn = HTTPRequestPart
+    public typealias InboundOut = Never
+
+    public typealias OutboundIn = Never
+    public typealias OutboundOut = HTTPResponsePart
+
+    var upload: HTTPResumableUpload
```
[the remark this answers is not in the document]
guoye-zhang: It is currently used in HTTPResumableUpload class. We could look into refactoring it.

Sources/NIOResumableUpload/HTTPResumableUploadChannel.swift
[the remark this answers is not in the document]
FranzBusch: We can also keep the force-unwraps but we should at least document why they are okay.

Sources/NIOResumableUpload/HTTPResumableUploadHandler.swift
FranzBusch: I think at least we should rename it to `channelInitializer`. I am still a bit worried about diverting from the norm here but if @Lukasa is happy with it we can go ahead.

guoye-zhang: Yeah, we could change that. The terminology `configurator` came from nio-http2: https://github.com/apple/swift-nio-http2/blob/a6c66bcbe7dfba6f23bff8599376b5fad858a818/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift#L336

Sources/NIOResumableUpload/HTTPResumableUpload.swift
```diff
@@ -0,0 +1,401 @@
...
+            handler.writeAndFlush(.head(informationalResponse), promise: nil)
+
+            let strippedRequest = HTTPResumableUploadProtocol.stripRequest(request)
+            let channel = self.createChannel(handler: handler, parent: channel)
+            channel.receive(.head(strippedRequest))
+        case .offsetRetrieving:
+            if let path = request.path, let upload = self.context.findUpload(path: path) {
+                self.uploadHandler = nil
+                self.uploadHandlerChannel.withLockedValue { $0 = nil }
+                upload.offsetRetrieving(otherHandler: handler)
```
jrflat, line 230: We immediately call `detachUploadHandler` in `offsetRetrieving` so I don't think we should set the handler or channel to `nil` here (otherwise `detachUploadHandler` does nothing).

Sources/NIOResumableUpload/HTTPResumableUploadHandler.swift
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {}
+
+    public func errorCaught(context: ChannelHandlerContext, error: Error) {
+        self.upload.end(handler: self, error: error)
+    }
+
+    public func read(context: ChannelHandlerContext) {
+        // Do nothing.
```
jrflat, line 98: Overriding this method to do nothing prevents upload resumption in HTTP/1.1 (the bug Marius was seeing), but I'm not entirely sure why. In the failing case, we do

```
HTTPResumableUpload offsetRetrieving(otherHandler:)
HTTPResumableUploadHandler channelRead(context:data:)
HTTPResumableUploadHandler channelReadComplete(context:)
HTTPResumableUploadHandler read(context:)
```

and the server hangs here while the client tries to send the PATCH request. If we remove this, the upload appending happens as expected right after offset retrieving. I'm still learning the NIO architecture, is there a reason we want to override the default `read(context:)` handling?

Lukasa: This seems unlikely to be what we want. `read` is used to ask the channel to read more from the network, so doing this will suppress all reads. Probably not what we want.

In general, empty overrides should be removed: their default behaviour is better.

guoye-zhang: There was a specific reason that we intercepted `read`. I think we wanted `read` to be driven by the child channel managed by the client's handlers, so this was the way to disable auto read behavior of the parent channel. Wondering if I missed something.

jrflat: That makes sense. Intercepting read like this does allow the child channel to drive the read behavior. But since the offset retrieving is transparent to the client's handler, I think the `read` doesn't propagate to the client's handler. I'm looking into how we could solve this while still overriding the `read`, maybe we need to explicitly attach the handler during offset retrieving?

jrflat: @Lukasa Guoye and I were debugging this and found that multiple H1 requests were being received on the same handler, which caused the issue since we were setting `self.uploadHandler` to `nil` (e.g. the subsequent upload appending `PATCH` wasn't being handled). Is there a way to ensure that only one request is delivered per handler?

Sources/NIOResumableUploadDemo/main.swift
```diff
@@ -0,0 +1,114 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
guoye-zhang, line 15: Unfortunately had to use Foundation to use `FileHandle` for cross-platform file IO

Sources/NIOResumableUploadDemo/main.swift
```diff
@@ -0,0 +1,114 @@
...
+        self.directory = directory
+    }
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        switch self.unwrapInboundIn(data) {
+        case .head(let request):
+            switch request.method {
+            case .post, .put:
+                if let requestPath = request.path {
+                    let path = self.directory.appending(requestPath)
```
guoye-zhang, line 44: This might be a bad idea since people can send `../` in the request path

guoye-zhang: I'm now using `standardized` URL and comparing the prefix which hopefully addresses the issue

PR #204, opened 2023-08-11
Enforce Sendability

gjcairo:
This PR fixes Sendability warnings present when enabling strict concurrency checking.

### Motivation:

We want to make sure `swift-nio-extras` is free of concurrency bugs, and thus removing Sendability warnings is crucial.

### Modifications:

Some types have been made Sendable, and in other places either patterns have been changed to ensure we're not crossing concurrency domains, or values have been wrapped with `NIOLoopBound` when we know we're on a specific EL.

### Result:

No more warnings in `swift-nio-extras`.

### DIFF
--- a/Sources/NIOExtras/PCAPRingBuffer.swift
+++ b/Sources/NIOExtras/PCAPRingBuffer.swift
@@ -12,18 +12,21 @@
 //
 //===----------------------------------------------------------------------===//
 
+import NIOConcurrencyHelpers
 import NIOCore
 
 // MARK: NIOPCAPRingBuffer
 /// Storage for the most recent set of packets captured subject to constraints.
 /// Use ``addFragment(_:)`` as the sink to a ``NIOWritePCAPHandler`` and call ``emitPCAP()``
 /// when you wish to get the recorded data.
-/// - Warning:  This class is not thread safe so should only be called from one thread.
-public class NIOPCAPRingBuffer {
-    private var pcapFragments: CircularBuffer<ByteBuffer>
-    private var pcapCurrentBytes: Int
+/// - Note: This class is annotated `@unchecked Sendable` because its mutable state is protected by a lock.
+public final class NIOPCAPRingBuffer: @unchecked Sendable {
     private let maximumFragments: Int
     private let maximumBytes: Int
+    
+    private let pcapLock = NIOLock()
+    private var pcapFragments: CircularBuffer<ByteBuffer>
+    private var pcapCurrentBytes: Int
 
     /// Initialise the buffer, setting constraints.
     /// - Parameters:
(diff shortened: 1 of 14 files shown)


Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -12,16 +12,17 @@
...
 
 // MARK: NIOPCAPRingBuffer
 /// Storage for the most recent set of packets captured subject to constraints.
 /// Use ``addFragment(_:)`` as the sink to a ``NIOWritePCAPHandler`` and call ``emitPCAP()``
 /// when you wish to get the recorded data.
 /// - Warning:  This class is not thread safe so should only be called from one thread.
-public class NIOPCAPRingBuffer {
-    private var pcapFragments: CircularBuffer<ByteBuffer>
-    private var pcapCurrentBytes: Int
+public final class NIOPCAPRingBuffer: Sendable {
```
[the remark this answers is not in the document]
gjcairo: This is a good question. This class is only used in tests by us (although it's public API).

I was trying to get rid of a warning where a `NIOPCAPRingBuffer` is captured in an `ELF/map` chained after a call to `write` on `ChannelHandlerContext` (the test in question is in `TriggerOnCumulativeSizeHandler/TriggerOnCumulativeSizeHandler/write`):

```swift
let boundedSink = NIOLoopBound(self.sink, eventLoop: context.eventLoop)
context.write(data).map { [pcapRingBuffer] in
  boundedSink.value(captureBytes(ringBuffer: pcapRingBuffer))
}.cascade(to: promise)
```

This piece of code also seems like a somewhat sensible pattern to do as a client when you're trying to debug something (which is why this class exists as public API, I assume). I'm not sure how to refactor this to get rid of the warning without making the class `Sendable`, as `pcapRingBuffer ` is modified as a result of the call to `context.write()`. This means I can't call `captureBytes(ringBuffer:)` outside of the closure, as we'd trigger the PCAP at the wrong time (causing the wrong behaviour and assertion failures).

I'm open to refactor suggestions if we want to avoid making this `Sendable`. Having said that, is there any reason why you would suggest avoiding it, Cory? Or would we be better off making these changes?

Sources/NIOExtras/PCAPRingBuffer.swift
```diff
@@ -12,18 +12,21 @@
...
-    private var pcapFragments: CircularBuffer<ByteBuffer>
-    private var pcapCurrentBytes: Int
+/// - Note: This class is annotated `@unchecked Sendable` because its mutable state is protected by a lock.
+public final class NIOPCAPRingBuffer: @unchecked Sendable {
     private let maximumFragments: Int
     private let maximumBytes: Int
+    
+    private let pcapLock = NIOLock()
+    private var pcapFragments: CircularBuffer<ByteBuffer>
+    private var pcapCurrentBytes: Int
```
glbrntt, line 29: Rather than using a lock should we create a `State` struct and stick that in a locked-value-box?

Sources/NIONFS3/NFSFileSystemInvoker.swift
```diff
@@ -39,122 +39,135 @@ internal struct NFS3FileSystemInvoker<FS: NFS3FileSystemNoAuth, Sink: NFS3FileSy
         case .mountNull:
             self.sink.sendSuccessfulReply(.mountNull, call: callMessage)
         case .mount(let call):
+            let boundedSink = NIOLoopBound(self.sink, eventLoop: self.eventLoop)
```
glbrntt, line 42: What guarantees that we're on the correct event-loop here?

Sources/NIOExtras/PCAPRingBuffer.swift
dnadoba: Yes, that would be better

PR #205, opened 2023-08-16
NFS3 demo server

weissi:
### Motivation:

Previously (#155), we added support for the NFS3 protocol. But without actually having a demo server, that's not actually very much fun.

### Modifications:

Add a demo server which can mount a simple dummy file system in user space.

### Result:

Better demo


Package.swift
```diff
@@ -135,6 +143,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.42.0"),
         .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
```
Lukasa, line 146: Ideally we'd avoid the swift-log dependency here. Any reason we can't `print`?

weissi: We use the logger to carry metadata through. So if you get an error in your file system, you can log and you'll get the precise connection etc associated. For production that's important and given that I had the code I left it here too because I thought the swift-log dependency wouldn't be too bad.

But I can take it out, lose the metadata propagation and just use `print(...)`

Lukasa: It's definitely not too bad, and I could be talked out of this, but in general keeping the dependency tree small is kinda nice. No strong feelings though, happy for someone to try to talk me out of it.

weissi: Similar position but the other way around. Happy either way really ;)

PR #208, opened 2023-10-06
Enable automatic compression format detection

dnadoba:
### Motivation:

Some servers can announce the wrong compression format. We should autodetect the format used through header detection of the compressed data.

### Modifications:

- enable `zlib`s automatic compression format detection

### Result:

We can decompress wrongly advertised requests and responses.


Lukasa (approved):
A bit late, but this looks good.

PR #209, opened 2023-11-02
Don't reserve capacity for NIOPCAPRingBuffer

glbrntt:
Motivation:

The NIOPCAPRingBuffer can limit the number of fragment or the total number of bytes in its buffer or both. When configuring the buffer to limit only the maximum number of bytes it sets the maximum number of fragments allowed to `.max`. On `init` the buffer reserves enough capacity to store the maximum number of fragments.

This would be a large and potentially totally unnecessary allocation. That is, if it didn't crash at runtime. It crashes at runtime as `CircularBuffer` converts the requested capacity to a `UInt32` which traps if you pass it an `Int.max`.

Modifications:

- Don't reserve capacity on init
- Adjust the test which tests the byte limit to not set a capacity as well

Result:

- `NIOPCAPRingBuffer(maximumBytes:)` doesn't crash

### DIFF
--- a/Sources/NIOExtras/PCAPRingBuffer.swift
+++ b/Sources/NIOExtras/PCAPRingBuffer.swift
@@ -35,7 +35,8 @@ public class NIOPCAPRingBuffer {
         self.maximumFragments = maximumFragments
         self.maximumBytes = maximumBytes
         self.pcapCurrentBytes = 0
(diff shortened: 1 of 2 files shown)


hamzahrmalik (approved):

PeterAdams-A (approved):
Seems sensible.

PR #211, opened 2023-11-28
Changes for Musl support.

al45tair:
Make sure we import `Musl` rather than `Glibc`.

### Motivation:

We want NIO to build against Musl.

### Modifications:

Added a smattering of `import Musl`s.

### Result:

swift-nio-extras will build against Musl.


FranzBusch:

FranzBusch:

Lukasa:

PR #212, opened 2023-11-28
Add retroactive marker to test conformances

Lukasa:
Motivation

Nightly CI builds require annotations on retroactive conformances.
We have a few used only in tests, which are totally safe.

Modifications

Add retroactive conformance marker.

Result

Nightly CI works again

### DIFF
--- a/Tests/NIOExtrasTests/DebugInboundEventsHandlerTest.swift
+++ b/Tests/NIOExtrasTests/DebugInboundEventsHandlerTest.swift
@@ -97,7 +97,7 @@ class DebugInboundEventsHandlerTest: XCTestCase {
     
 }
 
-extension DebugInboundEventsHandler.Event: Equatable {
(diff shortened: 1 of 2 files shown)


Tests/NIOExtrasTests/DebugInboundEventsHandlerTest.swift
```diff
@@ -123,3 +123,14 @@ extension DebugInboundEventsHandler.Event: Equatable {
...
+
+#if swift(>=5.8)
+#if $RetroactiveAttribute
+extension DebugInboundEventsHandler.Event: @retroactive Equatable { }
+#else
+extension DebugInboundEventsHandler.Event: Equatable { }
+#endif
+#else
+extension DebugInboundEventsHandler.Event: Equatable { }
+#endif
```
FranzBusch, line 135: Beautiful

FranzBusch (approved):

PR #214, opened 2023-12-15
Add missing availability guards in tests

simonjbeaumont:
## Motivation

Some of the test code was missing availability guards for Apple platforms, resulting in build failures for these platforms, e.g.

```
error: Concurrency is only available in iOS 13.0.0 or newer
```

## Modifications

Add missing availability guard to async function.

## Result

Tests can now build and run on iOS and other Apple platforms.


glbrntt (approved):

glbrntt:

PR #216, opened 2024-01-04
Make HTTP type conversion code public

adam-fowler:
Make NIOHTTP1 -> HTTPTypes and vice versa conversion code public

### Motivation:

WebSocket APIs still use NIOHTTP1 types which means any library that wants to use the new types has to write a conversion step. Also OpenAPI runtime uses new HTTP types while both the server transports that are published use old NIOHTTP1 types. It is probably better that there is one place where this conversion code lives.

### DIFF
--- a/Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
+++ b/Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
@@ -15,10 +15,23 @@
 import HTTPTypes
 import NIOHTTP1
 
-public enum HTTP1TypeConversionError: Error {
-    case invalidMethod
-    case missingPath
-    case invalidStatusCode
+public struct HTTP1TypeConversionError: Error, Equatable {
+    private enum Internal {
+        case invalidMethod
+        case missingPath
+        case invalidStatusCode
+    }
+    private let value: Internal
+    private init(_ value: Internal) {
+        self.value = value
+    }
+
+    /// Failed to create HTTPRequest.Method from HTTPMethod
+    public static var invalidMethod: Self { .init(.invalidMethod)}
+    /// Failed to extract a path from HTTPRequest
+    public static var missingPath: Self { .init(.missingPath)}
+    /// HTTPResponseHead had an invalid status code
+    public static var invalidStatusCode: Self { .init(.invalidStatusCode)}
 }
 
 extension HTTPMethod {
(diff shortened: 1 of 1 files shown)


Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
```diff
@@ -15,14 +15,14 @@
 import HTTPTypes
 import NIOHTTP1
 
-private enum HTTP1TypeConversionError: Error {
+public enum HTTP1TypeConversionError: Error {
```
FranzBusch, line 18: We have to do our usual struct wrapping enum dance here otherwise we can never evolve this enum in the future.

adam-fowler: Done.
I was wondering why NIO never implemented a `func ~` for these errors so switch and catch pattern matching was simplified. Currently to catch one of these "struct" errors you need
```swift
} catch let e as HTTP1TypeConversionError where e == .invalidStatusCode {
```
Which is a mouthful, and something I get wrong every time.
If this was added to allow pattern matching
```swift
    public static func ~= (_ lhs: Self, _ rhs: Error) -> Bool {
        guard let rhs = rhs as? Self else { return false }
        return lhs == rhs
    }
```
The catch statement gets reduced to
```swift
} catch HTTP1TypeConversionError.invalidStatusCode {
```
as if it is an enum.

FranzBusch: In general, we try to stay away from custom operators or overloading more operators since they have quite a heavy impact on the compiler's performance. What we really want is extensible enums in non-library evolution mode.

adam-fowler: > What we really want is extensible enums in non-library evolution mode.

Been waiting for that one for a while.

adam-fowler: @FranzBusch Is there anything holding merging this now?

FranzBusch (approved):
LGTM!

FranzBusch:

adam-fowler:
Is this ready to merge? The API breakage CI seems stuck.

FranzBusch:

FranzBusch:
@yim-lee for some reason the API breakage pipeline is not starting can you check what's going wrong here?

yim-lee:

yim-lee:
Just needed to add Adam to the allow list it seems. Should be ok now.

adam-fowler:
The API breakage is complaining about `LineBasedFrameDecoder.decode(buffer:) has been removed`

As I haven't touched this file I'm not sure why I'm getting this.

Issue #219, opened 2024-03-11
HTTPServerWithQuiescingDemo prints an error if the main code is wrapped in a function


crontab:
### Expected behavior
The demo should be able to shutdown gracefully on SIGINT.

### Actual behavior
If the top-level code is wrapped in a function like so:

```
func run() throws {
    // main top-level code, unchanged
}

try run()
```

then upon receiving SIGINT the app prints:

```
ERROR: Cannot schedule tasks on an EventLoop that has already shut down.
    This will be upgraded to a forced crash in future SwiftNIO versions.
```

and seems to shut down everything and exit normally.

This is a pretty big inconvenience as in any more or less complex project the code that bootstraps the server will most likely be in a function.

The main suspicion is that some destructor is called at the end of `run()` that tries to use an EventLoop. No idea which object in the main code that might be. Any help would be greatly appreciated!

### SwiftNIO-Extras version/commit hash

Current HEAD, _a33bb16e794432ffc1e6de49d2e7f9478689a6ac_

### Swift & OS version

Swift 5.10
macOS 14.3.1 (23D60)

crontab:
Oh, just found the solution, will post it here.

The function should isolate the `quiesce` object into a smaller block to ensure that its destructor is executed before the call to `group.syncShutdownGracefully()` like so:

```
func run() throws {
	let group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
	do {
		let quiesce = ServerQuiescingHelper(group: group)
		// ... rest of the code ...
		try fullyShutdownPromise.futureResult.wait()
	}
	try group.syncShutdownGracefully()
}
```

Finally, this doesn't print the error!

crontab:
P.S. may a suggest a modification to the demo app that fixes this?

Lukasa:
Absolutely, a patch would be welcomed that did that.

PR #220, opened 2024-03-16
Fixed the shutdown order for quiescing demo (#219)

crontab:
Fixes the problem with the shutdown order in HTTPServerWithQuiescingDemo.

### Motivation:

If the main server code in HTTPServerWithQuiescingDemo is isolated into a separate function (which will likely happen in any real-world project), a problem with the order in which various objects are created and destroyed becomes apparent: SIGINT will cause your project to print

```
ERROR: Cannot schedule tasks on an EventLoop that has already shut down.
    This will be upgraded to a forced crash in future SwiftNIO versions.
```

The reason for this is that the `quiesce` object's destructor tries to use the event loop after it's been shut down. The error is not printed in the original code likely because the destructors for top-level objects are not called at all (can't confirm this).

### Modifications:

1. The main server code is isolated into a function (`runServer()`) so that it's closer to real-world usage.
2. The code that creates `quiesce` and the rest is isolated into yet another nested block to ensure all objects are destroyed before the final call to `group.syncShutdownGracefully()`.

### Result:

The run-time error is no longer printed since the destructors are called in the right order.

### DIFF
--- a/Sources/HTTPServerWithQuiescingDemo/main.swift
+++ b/Sources/HTTPServerWithQuiescingDemo/main.swift
@@ -59,45 +59,54 @@ private final class HTTPHandler: ChannelInboundHandler {
     }
 }
 
-let group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
-let quiesce = ServerQuiescingHelper(group: group)
+private func runServer() throws {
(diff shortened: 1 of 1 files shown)


glbrntt (approved):
Thank you!

glbrntt:

PR #225, opened 2024-07-09
Conditional Response Compression

dimitribouniol:
Added support for conditional response compression in `HTTPResponseCompression`.

### Motivation:

This change allows libraries like vapor or direct users of the handler to add additional functionality around _when_ a response should be compressed, rather than always or never for the channel's lifetime. This is particularly important for pipelined requests, as some responses (ie. dynamic data) should be dynamically compressed, while others (ie. images) should not.

Two specific use cases immediately came to mind, particularly checking the content type of the response and using some heuristic, or using marker headers (ie. `X-Vapor-Compression`) to statelessly mark entire routes as compressible or incompressible by default.

### Modifications:

Rather than hardcode specific use cases, I went with an approach other handlers have used which is to provide a predicate to `HTTPResponseCompression` that allows the caller to weigh in on if compression should be performed. This allows a library like vapor to always install the handler, but then conditionally check the headers of the response about to go out to determine if it should be compressed or not based on heuristics and knobs it can choose to expose. Note that this predicate is always called even in cases where the request does not support compression, so the user of the handler does not need to concern themselves ahead of time whether or not to write marker headers based on the requests, and can keep the logic and cleanup contained in one place.

### Result:

`HTTPResponseCompression` now supports conditional response compression allowing for greater flexibility in how the handler is used, which can hopefully increase its general usefulness.

### DIFF
--- a/Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
+++ b/Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
@@ -78,13 +78,37 @@ public final class HTTPResponseCompressor: ChannelDuplexHandler, RemovableChanne
     ///
     /// - Parameter responseHeaders: The headers that will be used for the response. These can be modified as needed at this stage, to clean up any marker headers used to statelessly determine if compression should occur, and the new headers will be used when writing the response. Compression headers are not yet provided and should not be set; ``HTTPResponseCompressor`` will set them accordingly based on the result of this predicate.
     /// - Parameter isCompressionSupported: Set to `true` if the client requested compatible compression, and if the HTTP response supports it, otherwise `false`.
-    /// - Returns: Return `true` if the compressor should proceed to compress the response, or `false` if the response should not be compressed.
+    /// - Returns: Return ``CompressionIntent/compressIfPossible`` if the compressor should proceed to compress the response, or ``CompressionIntent/doNotCompress`` if the response should not be compressed.
     ///
-    /// - Note: Returning `true` when compression is not supported will not enable compression, and the modified headers will always be used.
+    /// - Note: Returning ``CompressionIntent/compressIfPossible`` is only a suggestion — when compression is not supported, the response will be returned as is along with any modified headers.
     public typealias ResponseCompressionPredicate = (
         _ responseHeaders: inout HTTPResponseHead,
         _ isCompressionSupported: Bool
-    ) -> Bool
+    ) -> CompressionIntent
+    
+    /// A signal a ``ResponseCompressionPredicate`` returns to indicate if it intends for compression to be used or not when supported by HTTP.
+    public struct CompressionIntent: Sendable, Hashable {
+        /// The internal type ``CompressionIntent`` uses.
+        enum RawValue {
+            /// The response should be compressed if supported by the HTTP protocol.
+            case compressIfPossible
+            /// The response should not be compressed even if supported by the HTTP protocol.
+            case doNotCompress
+        }
+        
+        /// The raw value of the intent.
+        let rawValue: RawValue
+        
+        /// Initialize the raw value with an internal intent.
+        init(_ rawValue: RawValue) {
+            self.rawValue = rawValue
+        }
+        
+        /// The response should be compressed if supported by the HTTP protocol.
+        public static let compressIfPossible = CompressionIntent(.compressIfPossible)
+        /// The response should not be compressed even if supported by the HTTP protocol.
+        public static let doNotCompress = CompressionIntent(.doNotCompress)
+    }
 
     /// Errors which can occur when compressing
     public enum CompressionError: Error {
@@ -152,10 +176,10 @@ public final class HTTPResponseCompressor: ChannelDuplexHandler, RemovableChanne
             let requestSupportsCompression = algorithm != nil && responseHead.status.mayHaveResponseBody
             
             /// If a predicate was set, ask it if we should compress when compression is supported, and give the predicate a chance to clean up any marker headers that may have been set even if compression were not supported.
-            let predicateSupportsCompression = responseCompressionPredicate?(&responseHead, requestSupportsCompression) ?? true
+            let predicateCompressionIntent = responseCompressionPredicate?(&responseHead, requestSupportsCompression) ?? .compressIfPossible
             
             /// Make sure that compression should proceed, otherwise stop here and supply the response headers before configuring the compressor.
-            guard let algorithm, requestSupportsCompression, predicateSupportsCompression else {
+            guard let algorithm, requestSupportsCompression, predicateCompressionIntent == .compressIfPossible else {
                 context.write(wrapOutboundOut(.head(responseHead)), promise: promise)
                 return
             }
--- a/Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
+++ b/Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
@@ -57,6 +57,13 @@ private func qValueFromHeader<S: StringProtocol>(_ text: S) -> Float {
 /// ahead-of-time instead of dynamically, could be a waste of CPU time and latency for relatively minimal
 /// benefit. This channel handler should be present in the pipeline only for dynamically-generated and
 /// highly-compressible content, which will see the biggest benefits from streaming compression.
+///
+/// The compressor optionally accepts a predicate to help it determine on a per-request basis if compression
+/// should be used, even if the client requests it for the request. This could be used to conditionally and statelessly
+/// enable compression based on resource types, or by emitting and checking for marker headers as needed.
+/// Since the predicate is always called, it can also be used to clean up those marker headers if compression was
+/// not actually supported for any reason (ie. the client didn't provide compatible `Accept` headers, or the
+/// response was missing a body due to a special status code being used)
 public final class HTTPResponseCompressor: ChannelDuplexHandler, RemovableChannelHandler {
     /// This class accepts `HTTPServerRequestPart` inbound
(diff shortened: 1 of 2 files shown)


Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
```diff
@@ -84,11 +103,23 @@ public final class HTTPResponseCompressor: ChannelDuplexHandler, RemovableChanne
...
     private let initialByteBufferCapacity: Int
+    private let responseCompressionPredicate: ResponseCompressionPredicate?
 
-    /// Initialise a ``HTTPResponseCompressor``
+    /// Initialize a ``HTTPResponseCompressor``.
+    /// - Parameter initialByteBufferCapacity: Initial size of buffer to allocate when hander is first added.
+    @_disfavoredOverload
+    @available(*, deprecated, message: "Deprecated in favor of HTTPResponseCompressor(initialByteBufferCapacity:, responseCompressionPredicate:)")
+    public convenience init(initialByteBufferCapacity: Int = 1024) {
+        // TODO: This version is kept around for backwards compatibility and should be merged with the signature below in the next major version.
```
dimitribouniol, line 113: Not sure what the best practice is here for deprecating overloads that can break backwards compatibility. I remember this was a thing early on in Swift 5, but not sure if its still relevant?

dimitribouniol:

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
Lukasa: Let's not deprecate this: we can just have two methods and remove the default argument for the new one.

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
```diff
@@ -84,11 +103,23 @@ public final class HTTPResponseCompressor: ChannelDuplexHandler, RemovableChanne
...
+        self.init(initialByteBufferCapacity: initialByteBufferCapacity, responseCompressionPredicate: nil)
+    }
+    
+    /// Initialize a ``HTTPResponseCompressor``.
     /// - Parameter initialByteBufferCapacity: Initial size of buffer to allocate when hander is first added.
-    public init(initialByteBufferCapacity: Int = 1024) {
+    /// - Parameter responseCompressionPredicate: The predicate used to determine if the response should be compressed or not based on its headers. Defaults to `nil`, which will compress every response this handler sees. This predicate is always called wether the client supports compression for this response or not, so it can be used to clean up any marker headers you may use to determine if compression should be performed or not. Please see ``ResponseCompressionPredicate`` for more details.
```
Lukasa, line 119:
```suggestion
    /// - Parameter responseCompressionPredicate: The predicate used to determine if the response should be compressed or not based on its headers. Defaults to `nil`, which will compress every response this handler sees. This predicate is always called whether the client supports compression for this response or not, so it can be used to clean up any marker headers you may use to determine if compression should be performed or not. Please see ``ResponseCompressionPredicate`` for more details.
```

Tests/NIOHTTPCompressionTests/HTTPResponseCompressorTest.swift
```diff
@@ -688,6 +724,234 @@ class HTTPResponseCompressorTest: XCTestCase {
             }
         }
     }
+    
+    func testConditionalCompressionEnabled() throws {
+        let predicateWasCalled = expectation(description: "Predicate was called")
+        let compressor = HTTPResponseCompressor { responseHeaders, isCompressionSupported in
+            predicateWasCalled.fulfill()
```
Lukasa, line 731: We should probably fulfil the predicate last.

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
```diff
@@ -66,6 +73,18 @@ public final class HTTPResponseCompressor: ChannelDuplexHandler, RemovableChanne
...
     /// This class emits `HTTPServerResponsePart` outbound.
     public typealias OutboundOut = HTTPServerResponsePart
+    
+    /// A closure that accepts a response header, optionally modifies it, and returns `true` if the response it belongs to should be compressed.
+    ///
+    /// - Parameter responseHeaders: The headers that will be used for the response. These can be modified as needed at this stage, to clean up any marker headers used to statelessly determine if compression should occur, and the new headers will be used when writing the response. Compression headers are not yet provided and should not be set; ``HTTPResponseCompressor`` will set them accordingly based on the result of this predicate.
+    /// - Parameter isCompressionSupported: Set to `true` if the client requested compatible compression, and if the HTTP response supports it, otherwise `false`.
+    /// - Returns: Return `true` if the compressor should proceed to compress the response, or `false` if the response should not be compressed.
+    ///
+    /// - Note: Returning `true` when compression is not supported will not enable compression, and the modified headers will always be used.
```
Lukasa, line 83: This behaviour is surprising to me. I think we should introduce a new enum for the return value, something like `compressIfPossible` and `doNotCompress` as the cases. Just to clarify what will happen.

Lukasa (commented):
Generally looks really nice. Just a few nits here.

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
dimitribouniol: I was thinking about this too, glad we are on the same page. Are we ok with `enum` enums, or should I make a struct with private enums for future compatibility?

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
dimitribouniol: Should I keep the todo? Is there another mechanism to not forget these on the next major version bump?

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
dimitribouniol: Went with a struct for futureproofness after all, but let me know if the cases are clearly named (went with an "intent"-like terminology, and clarified the wording in the doc comment)

Tests/NIOHTTPCompressionTests/HTTPResponseCompressorTest.swift
dimitribouniol: Put them inside a defer instead so this is guaranteed, unless that would lead to a bad outcome? Its only really trying to make sure the predicate isn't silently skipped.

Lukasa: Defer is just fine.

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
Lukasa: Please go ahead and file an issue and we can mark it for the next major version, but I'm happy to keep the TODO as well.

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
```diff
@@ -66,6 +73,42 @@ public final class HTTPResponseCompressor: ChannelDuplexHandler, RemovableChanne
...
+        
+        /// Initialize the raw value with an internal intent.
+        init(_ rawValue: RawValue) {
+            self.rawValue = rawValue
+        }
+        
+        /// The response should be compressed if supported by the HTTP protocol.
+        public static let compressibleResponse = CompressionIntent(.compressibleResponse)
+        /// The response should not be compressed even if supported by the HTTP protocol.
+        public static let uncompressedResponse = CompressionIntent(.uncompressedResponse)
```
Lukasa, line 110: I think let's make it clear these are action requests:

```
.compressIfPossible
.doNotCompress
```

dimitribouniol: Updated!

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
dimitribouniol: Linked in the code as well: https://github.com/apple/swift-nio-extras/issues/226

Lukasa:

Sources/NIOHTTPCompression/HTTPResponseCompressor.swift
```diff
@@ -84,11 +127,21 @@ public final class HTTPResponseCompressor: ChannelDuplexHandler, RemovableChanne
     private var pendingWritePromise: EventLoopPromise<Void>!
 
     private let initialByteBufferCapacity: Int
+    private let responseCompressionPredicate: ResponseCompressionPredicate?
 
-    /// Initialise a ``HTTPResponseCompressor``
+    /// Initialize a ``HTTPResponseCompressor``.
+    /// - Parameter initialByteBufferCapacity: Initial size of buffer to allocate when hander is first added.
+    public convenience init(initialByteBufferCapacity: Int) {
```
Lukasa, line 134: Hmm, when I suggested taking off the default argument value I actually meant to take it from the other initializer's `responseCompressionPredicate`. But I now wonder: why can't we have default values here? There should be no ambiguity, the compiler should be fine.

dimitribouniol: Initially I ran into trouble, but I forgot what the issue was — let me give it another try

dimitribouniol: You're right, we can have default arguments for both of them after all, let me add it back

Lukasa (approved):
I love this, this is fantastic work. Thanks @dimitribouniol!

PR #228, opened 2024-09-09
Add HTTPResponseCompressor intent for cases where requested compression algorithm is not supported

adincebic:
Adds a new HTTPResponseCompressor intent that should be used to signal when compression is not supported.

### Motivation:

Vapor currently does not return a 415 status code when compression is not supported. This change will help to make the code less ambiguous when we start supporting that use case.

### Modifications:

I added a new case to the raw type for the compression intent `doNotCompressWithUnsupportedAlgorithm` and updated the docs.

### Result:

Behaviourally nothing changes except for the fact that clients can express their intent more clearly.


FranzBusch (commented):
Can we add tests showing how this is used?

PR #241, opened 2024-12-16
support h2 stream resets through user events

ehaydenr:
Allow applications to trigger HTTP/2 stream resets while using NIOHTTPTypesHTTP2's codecs

### Motivation:

Resetting streams with specific error codes is required by some applications such as those implementing the CONNECT method
(see https://datatracker.ietf.org/doc/html/rfc9113#section-8.5-8). Unfortunately, the HTTP2ToHTTP codecs don't expose this capability to applications.

### Modifications:

Introduce an outbound user event applications can trigger when needing to reset an HTTP/2 stream.

### Result:

Now applications can trigger HTTP/2 stream resets while using the codecs provided by NIOHTTPTypesHTTP2


glbrntt (approved):
LGTM, thanks Eric!

PR #242, opened 2024-12-17
Responsiveness under Working Conditions

ehaydenr:
Implementation of https://datatracker.ietf.org/doc/draft-ietf-ippm-responsiveness/ (draft 5) with flexible download and upload handlers to suit other use cases as well.

### Motivation:

The provided handlers are useful for measuring responsiveness and testing things like performance of proxies

### Modifications:

Add `NIOHTTPResponsiveness` and `NIOHTTPResponsivenessServer`

### Result:

We'll now have an implementation of the Responsiveness under Working Conditions draft


Sources/NIOHTTPResponsiveness/HTTPDrippingDownloadHandler.swift
```diff
@@ -0,0 +1,228 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import HTTPTypes
+import NIOCore
+import NIOHTTPTypes
+
+/// HTTP request handler sending a configurable stream of zeroes
```
Lukasa, line 19: Can we document that this requires the use of the HTTPTypes types, not the regular NIO ones.

Sources/NIOHTTPResponsiveness/ResponsivenessConfig.swift
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public struct ResponsivenessConfig: Encodable {
```
Lukasa, line 15: We should add the extra protocols to all of these. `Codable`, `Hashable, `Sendable`.

Sources/NIOHTTPResponsiveness/HTTPDrippingDownloadHandler.swift
Lukasa: Can I suggest that, for now, we don't make the two "helper" handlers `public`? It's generally wise for us not to commit to too much API without a clear use-case. Making something `public` that was previously `internal` is generally easy, changing something that we accidentally made `public` is much harder.

ehaydenr: The two "helper" handlers you're referring to are the HTTPDrippingDownloadHandler and the HTTPReceiveDiscardHandler, right? The idea being only the ResponsivenessConfig and SimpleResponsivenessRequestMux end up being public?

I think doing that would end up making this more of an example than a building block. I find that request multiplexers are typically not very reusable (hence the prefix "Simple"). My expectation is that someone wanting to put together a robust implementation would have their own request mux and add things like limits on download size, delay durations, etc.

glbrntt (changes requested):

Sources/NIOHTTPResponsiveness/SimpleResponsivenessRequestMux.swift
```diff
@@ -0,0 +1,141 @@
...
+                uriComponentIterator.next(), uriComponentIterator.next().flatMap { Int($0) }
+            ) {
+            case (.get, .some(""), .some("responsiveness"), .none, .none):
+                self.writeSimpleResponse(
+                    context: context,
+                    status: .ok,
+                    body: self.responsivenessConfigBuffer
+                )
+            case (.get, .some(""), .some("responsiveness"), .some("download"), .some(let count)):
+                self.addHandlerOrInternalError(context: context, handler: HTTPDrippingDownloadHandler(count: 1, size: count))
```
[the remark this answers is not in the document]
ehaydenr: The ordering is correct. a single chunk of size N.The variable and parameter naming, however, is very confusing. Updated to try to be less confusing

Sources/NIOHTTPResponsiveness/HTTPDrippingDownloadHandler.swift
```diff
@@ -0,0 +1,228 @@
...
+                currentChunkBytesLeft: self.size
+            )
+        )
+
+        self.writeChunk(context: context)
+    }
+
+    public func channelInactive(context: ChannelHandlerContext) {
+        self.phase = .done
+        self.scheduled?.cancel()
```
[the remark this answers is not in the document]
ehaydenr: done!

Sources/NIOHTTPResponsiveness/HTTPDrippingDownloadHandler.swift
```diff
@@ -0,0 +1,228 @@
...
+                at: HTTPDrippingDownloadHandler.downloadBodyChunk.readerIndex,
+                length: toSend
+            )!
+            context.write(self.wrapOutboundOut(.body(buffer)), promise: nil)
+            drippingState.currentChunkBytesLeft -= toSend
+            dataWritten = true
+        }
+        if dataWritten {
+            context.flush()
+        }
```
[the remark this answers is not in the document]
ehaydenr: Had to be careful here and put an extra `if dataWritten` inside one of the early returns, but the new code does avoid the unnecessary flush

Sources/NIOHTTPResponsiveness/HTTPDrippingDownloadHandler.swift
```diff
@@ -0,0 +1,228 @@
...
+            context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
+            return
+        }
+
+        // More chunks to write.. Kick off timer
+        drippingState.currentChunkBytesLeft = self.size
+        self.phase = .dripping(drippingState)
+        let this = NIOLoopBound(self, eventLoop: context.eventLoop)
+        let loopBoundContext = NIOLoopBound(context, eventLoop: context.eventLoop)
+        self.scheduled = context.eventLoop.scheduleTask(in: self.frequency) {
```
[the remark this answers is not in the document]
ehaydenr: Thanks for the suggestion. I _think_ I used it mostly correctly, but I'm not 100% sure. I haven't built much intuition around when swift does or doesn't allocate. I did find the fact that `scheduleCallback` throws a bit surprising and looked for an equivalent syncOperations-like method to avoid the `try`, but I didn't find one. I think `try!` may be fine in practice given current implementations, but not entirely correct. Interested in your feedback here.

Sources/NIOHTTPResponsiveness/HTTPReceiveDiscardHandler.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private func onExpectationViolated(context: ChannelHandlerContext, expectation: Int) {
+        self.expectationViolated = true
+
+        let body = ByteBuffer(
+            string:
+                "Received in excess of expectation; expected(\(expectation)) received(\(self.received))"
+        )
+        self.writeSimpleResponse(context: context, status: .ok, body: body)
```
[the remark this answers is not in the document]
ehaydenr: Yikes! Fixed, thanks

Sources/NIOHTTPResponsiveness/HTTPDrippingDownloadHandler.swift
```diff
@@ -0,0 +1,246 @@
...
+        // More chunks to write.. Kick off timer
+        drippingState.currentChunkBytesLeft = self.size
+        self.phase = .dripping(drippingState)
+        if self.scheduledCallbackHandler == nil {
+            let this = NIOLoopBound(self, eventLoop: context.eventLoop)
+            let loopBoundContext = NIOLoopBound(context, eventLoop: context.eventLoop)
+            self.scheduledCallbackHandler = HTTPDrippingDownloadHandlerScheduledCallbackHandler(handler: this, context: loopBoundContext)
+        }
+        // SAFTEY: scheduling the callback only potentially throws when invoked off eventloop
+        try! context.eventLoop.scheduleCallback(in: self.frequency, handler: self.scheduledCallbackHandler!)
```
glbrntt, line 234: Different EL implementations could throw for other reasons. In practice that _probably_ won't be the case but it'd be safer to catch the error and call `context.fireErrorCaught` instead of `try!` here.

ehaydenr: Good point - fixed

glbrntt (approved):

glbrntt:
@Lukasa did you also want to review as you left a few comments before?

Lukasa (approved):
I think this looks great, let's land it.

Lukasa:
Apologies for the delay here @ehaydenr, mind fixing up the conflicts for me?

PR #248, opened 2025-02-13
Eliminate hasPrefix usage for pseudo HTTP header names

guoye-zhang:
UTF8 view is faster than hasPrefix

rdar://144395951


glbrntt (approved):
Awesome, thanks @guoye-zhang!

Lukasa (approved):

ekinnear (approved):

PR #253, opened 2025-03-20
Add AsynchronizedFileSink for non-blocking PCAP recording [WIP]

RaghavRoy145:
### Motivation:

The current PCAP recording functionality uses SynchronizedFileSink, which blocks the event loop, this patch attempts an introduction of AsynchronizedFileSink,which uses Swift's concurrency model to perform non-blocking file writes. This patch attempts to address https://github.com/swift-server/async-http-client/issues/239.

This change also ensures that debugging via PCAP recording remains accessible while preventing the event loop from stalling. Modifications:

    - Implemented AsynchronizedFileSink thread-safe, non-blocking writes.
    - Provided unit tests to verify correct behavior.

### Result:

    - Users can enable non-blocking PCAP recording using AsynchronizedFileSink, improving debugging without risking connection pool performance.
    - Test coverage ensures the new implementation works as expected.

### WIP:

I have marked this PR WIP as the test I have written does not work properly and I was looking for any help with this from the community. The .pcap files get written however the test ends up never exiting, possible due to file close or flush not implemented correctly. I'm new to swift and find Swift on Server really cool, so my design decisions could be dubious! This issue seemed like a great way to get my hands dirty with swift concurrency, though I'm not sure if its the best way to solve the async PCAP recording problem as mentioned in the issue.

### DIFF
--- a/Sources/NIOExtras/WritePCAPHandler.swift
+++ b/Sources/NIOExtras/WritePCAPHandler.swift
@@ -808,6 +808,131 @@ extension NIOWritePCAPHandler {
             }
         }
     }
+
+    public final class AsynchronizedFileSink {
+        private let fileHandle: NIOFileHandle
+        private let eventLoop: EventLoop
+        private let errorHandler: (Swift.Error) -> Void
+        private var state: State = .running
+        
+        public enum FileWritingMode {
+            case appendToExistingPCAPFile
+            case createNewPCAPFile
+        }
+        
+        public struct Error: Swift.Error {
+            public var errorCode: Int
+
+            internal enum ErrorCode: Int {
+                case cannotOpenFileError = 1
+                case cannotWriteToFileError
+            }
+        }
+        
+        private enum State {
+            case running
+            case error(Swift.Error)
+        }
+        
+        /// Creates an AsynchronizedFileSink for writing to a .pcap file at the given path.
+        /// If fileWritingMode is `.createNewPCAPFile`, a file header is written.
+        public static func fileSinkWritingToFile(
+            path: String,
+            fileWritingMode: FileWritingMode = .createNewPCAPFile,
+            errorHandler: @escaping (Swift.Error) -> Void,
+            on eventLoop: EventLoop
+        ) async throws -> AsynchronizedFileSink {
+            let oflag: CInt = fileWritingMode == .createNewPCAPFile ? (O_TRUNC | O_CREAT) : O_APPEND
+            let fd: CInt = try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<CInt, Swift.Error>) in
+                path.withCString { pathPtr in
+                    let fd: Int32 = open(pathPtr, O_WRONLY | oflag, 0o600)
+                    if fd < 0 {
+                        continuation.resume(throwing: Error(errorCode: Error.ErrorCode.cannotOpenFileError.rawValue))
+                    } else {
+                        continuation.resume(returning: fd)
+                    }
+                }
+            }
+            
+            if fileWritingMode == .createNewPCAPFile {
+                _ = try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<CInt, Swift.Error>) in
+                    NIOWritePCAPHandler.pcapFileHeader.withUnsafeReadableBytes { ptr in
+                        let writeOk: Bool = sysWrite(fd, ptr.baseAddress, ptr.count) == ptr.count
+                        if writeOk {
+                            continuation.resume(returning: 0)  // Dummy value
+                        } else {
+                            continuation.resume(throwing: Error(errorCode: Error.ErrorCode.cannotWriteToFileError.rawValue))
+                        }
+                    }
+                }
+            }
+            let fileHandle: NIOFileHandle = NIOFileHandle(_deprecatedTakingOwnershipOfDescriptor: fd)
+            return AsynchronizedFileSink(fileHandle: fileHandle, eventLoop: eventLoop, errorHandler: errorHandler)
+        }
+        
+        private init(
+            fileHandle: NIOFileHandle,
+            eventLoop: EventLoop,
+            errorHandler: @escaping (Swift.Error) -> Void
+        ) {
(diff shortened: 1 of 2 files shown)


Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -808,6 +808,131 @@ extension NIOWritePCAPHandler {
...
+        /// Creates an AsynchronizedFileSink for writing to a .pcap file at the given path.
+        /// If fileWritingMode is `.createNewPCAPFile`, a file header is written.
+        public static func fileSinkWritingToFile(
+            path: String,
+            fileWritingMode: FileWritingMode = .createNewPCAPFile,
+            errorHandler: @escaping (Swift.Error) -> Void,
+            on eventLoop: EventLoop
+        ) async throws -> AsynchronizedFileSink {
+            let oflag: CInt = fileWritingMode == .createNewPCAPFile ? (O_TRUNC | O_CREAT) : O_APPEND
+            let fd: CInt = try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<CInt, Swift.Error>) in
```
Lukasa, line 846: I don't think this continuation is needed: the operations in the block are strictly synchronous.

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -808,6 +808,131 @@ extension NIOWritePCAPHandler {
...
+                    if fd < 0 {
+                        continuation.resume(throwing: Error(errorCode: Error.ErrorCode.cannotOpenFileError.rawValue))
+                    } else {
+                        continuation.resume(returning: fd)
+                    }
+                }
+            }
+            
+            if fileWritingMode == .createNewPCAPFile {
+                _ = try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<CInt, Swift.Error>) in
```
Lukasa, line 858: Same note here, this isn't actually asynchronous and doesn't need a continuation.

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -808,6 +808,131 @@ extension NIOWritePCAPHandler {
...
+            eventLoop: EventLoop,
+            errorHandler: @escaping (Swift.Error) -> Void
+        ) {
+            self.fileHandle = fileHandle
+            self.eventLoop = eventLoop
+            self.errorHandler = errorHandler
+        }
+        
+        public func write(buffer: ByteBuffer) async throws {
+            try await self.eventLoop.submit {
```
Lukasa, line 884: This still blocks the event loop. By being `async`, this method will already run off the event loop, so we don't need to do anything other than issue the write call.

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -808,6 +808,131 @@ extension NIOWritePCAPHandler {
...
+                            return written
+                        }
+                    }
+                }
+            }.get()
+        }
+        
+        /// Asynchronously syncs the file to disk using fsync.
+        public func asyncSync() async throws {
+            try await withCheckedThrowingContinuation { continuation in
```
Lukasa, line 902: Same note here, no need for the `submit`.

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -808,6 +808,131 @@ extension NIOWritePCAPHandler {
...
+                    } catch {
+                        continuation.resume(throwing: error)
+                    }
+                }
+            }
+        }
+
+        /// Asynchronously closes the file sink.
+        public func close() async throws {
+            try await withCheckedThrowingContinuation { continuation in
```
Lukasa, line 921: Same note here, no need for the submit.

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -808,6 +808,131 @@ extension NIOWritePCAPHandler {
...
+                        print("Error closing file: \(error)")
+                    }
+                }
+            }
+        }
+    }
 }
 
 extension NIOWritePCAPHandler.SynchronizedFileSink: @unchecked Sendable {}
+extension NIOWritePCAPHandler.AsynchronizedFileSink: @unchecked Sendable {}
```
Lukasa, line 938: Let's put the state in a `NIOLockedValueBox`. Then we can make this checked Sendable.

Sources/NIOExtras/WritePCAPHandler.swift
```diff
@@ -808,6 +808,131 @@ extension NIOWritePCAPHandler {
             }
         }
     }
+
+    public final class AsynchronizedFileSink {
```
Lukasa, line 812: I think a better version of this uses _NIOFileSystem to provide the async file I/O APIs. That will also avoid the unsafe code. @glbrntt how do we feel about taking that dependency here?

Sources/NIOExtras/WritePCAPHandler.swift
RaghavRoy145: Thanks for all your reviews! I've tried to take care of this in the new commit

Sources/NIOExtras/WritePCAPHandler.swift
RaghavRoy145: Got it!

Sources/NIOExtras/WritePCAPHandler.swift
RaghavRoy145: Oh I didn't consider this, mostly took inspiration from the implementation of SynchronizedFileSink, though I don't think using the filesystem instead would be too tough. Let me know!

RaghavRoy145: Oh I didn't consider this, mostly took inspiration from the implementation of SynchronizedFileSink, though I don't think using the filesystem instead would be too tough. Let me know!

glbrntt: I'm a little hesitant to do it before we've made the `FilePath` changes as that may well have unavoidable API breaks.

PR #254, opened 2025-03-27
Strict concurrency for NIOExtras and NIOExtrasTests

glbrntt:


fabianfett (approved):

Sources/NIOExtras/RequestResponseHandler.swift
```diff
@@ -88,7 +88,18 @@ public final class RequestResponseHandler<Request, Response>: ChannelDuplexHandl
...
+        // If the event loop of the promise is the same as the context then there's no
+        // change in isolation. Otherwise transfer the response onto the correct event-loop
+        // before succeeding the promise.
+        if promise.futureResult.eventLoop === context.eventLoop {
+            promise.assumeIsolatedUnsafeUnchecked().succeed(response)
+        } else {
+            let unsafeTransfer = UnsafeTransfer(response)
+            promise.futureResult.eventLoop.execute {
+                let response = unsafeTransfer.wrappedValue
+                promise.assumeIsolatedUnsafeUnchecked().succeed(response)
```
Lukasa, line 100: This pattern looks very dodgy. I suspect this reflects a very real problem with the API of this type. Did we file an issue for this?

glbrntt: Agreed, I wasn't happy with this but it was the least bad option I came up with at the time. I just filed https://github.com/apple/swift-nio-extras/issues/267

PR #266, opened 2025-04-08
Ignore return value of signal() on Android

marcprux:
See https://github.com/apple/swift-nio/pull/3181 and https://github.com/apple/swift-nio/issues/3180


Lukasa (approved):
Nice one, thanks!

PR #269, opened 2025-04-30
Add a `TimedCertificateReloader`

gjcairo:
This PR adds a `CertificateReloader` protocol and a `TimedCertificateReloader` implementation.

A `CertificateReloader` allows for certificate-key pairs to be observed and updated. A `TimedCertificateReloader` does this every set interval of time.

This PR also provides an extension on `TLSConfiguration` to enable this reloading ability in NIO applications, so that updated certificates and keys can be used during SSL handshakes; as well as conformance of the `TimedCertificateReloader` to `swift-service-lifecycle`'s `Service` protocol, for easy composition.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -258,7 +258,7 @@ var targets: [PackageDescription.Target] = [
         swiftSettings: strictConcurrencySettings
     ),
     .target(
-        name: "NIOCertificateHotReloading",
+        name: "NIOCertificateReloading",
         dependencies: [
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
@@ -269,9 +269,9 @@ var targets: [PackageDescription.Target] = [
         swiftSettings: strictConcurrencySettings
     ),
     .testTarget(
-        name: "NIOCertificateHotReloadingTests",
+        name: "NIOCertificateReloadingTests",
         dependencies: [
-            "NIOCertificateHotReloading",
+            "NIOCertificateReloading",
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "X509", package: "swift-certificates"),
@@ -292,7 +292,7 @@ let package = Package(
         .library(name: "NIOHTTPTypesHTTP2", targets: ["NIOHTTPTypesHTTP2"]),
         .library(name: "NIOResumableUpload", targets: ["NIOResumableUpload"]),
         .library(name: "NIOHTTPResponsiveness", targets: ["NIOHTTPResponsiveness"]),
-        .library(name: "NIOCertificateHotReloading", targets: ["NIOCertificateHotReloading"]),
+        .library(name: "NIOCertificateReloading", targets: ["NIOCertificateReloading"]),
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.81.0"),
--- a/Package.swift
+++ b/Package.swift
@@ -263,7 +263,7 @@ var targets: [PackageDescription.Target] = [
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "X509", package: "swift-certificates"),
-            .product(name: "ServiceLifecycle", package: "swift-service-lifecycle"),
+            .product(name: "SwiftASN1", package: "swift-asn1")
         ],
         swiftSettings: strictConcurrencySettings
     ),
@@ -274,6 +274,7 @@ var targets: [PackageDescription.Target] = [
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "X509", package: "swift-certificates"),
+            .product(name: "SwiftASN1", package: "swift-asn1")
         ],
         swiftSettings: strictConcurrencySettings
     ),
@@ -301,7 +302,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-algorithms.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-certificates.git", branch: "1.10.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.29.3"),
-        .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.8.0"),
+        .package(url: "https://github.com/apple/swift-asn1.git", from: "1.3.1"),
     ],
     targets: targets
 )
--- a/Package.swift
+++ b/Package.swift
@@ -303,7 +303,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-http-structured-headers.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-algorithms.git", from: "1.2.0"),
-        .package(url: "https://github.com/apple/swift-certificates.git", branch: "1.10.0"),
+        .package(url: "https://github.com/apple/swift-certificates.git", from: "1.10.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.29.3"),
         .package(url: "https://github.com/apple/swift-asn1.git", from: "1.3.1"),
         .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.8.0"),
--- a/Package.swift
+++ b/Package.swift
@@ -263,6 +263,7 @@ var targets: [PackageDescription.Target] = [
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "X509", package: "swift-certificates"),
+            .product(name: "ServiceLifecycle", package: "swift-service-lifecycle"),
         ],
         swiftSettings: strictConcurrencySettings
     ),
@@ -300,6 +301,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-algorithms.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-certificates.git", branch: "1.10.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.29.3"),
+        .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.8.0"),
     ],
     targets: targets
 )
--- a/Package.swift
+++ b/Package.swift
@@ -263,7 +263,6 @@ var targets: [PackageDescription.Target] = [
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "X509", package: "swift-certificates"),
-            .product(name: "SwiftASN1", package: "swift-asn1")
         ],
         swiftSettings: strictConcurrencySettings
     ),
@@ -274,7 +273,6 @@ var targets: [PackageDescription.Target] = [
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "X509", package: "swift-certificates"),
-            .product(name: "SwiftASN1", package: "swift-asn1")
         ],
         swiftSettings: strictConcurrencySettings
     ),
@@ -302,7 +300,6 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-algorithms.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-certificates.git", branch: "1.10.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.29.3"),
-        .package(url: "https://github.com/apple/swift-asn1.git", from: "1.3.1"),
     ],
     targets: targets
 )
--- a/Package.swift
+++ b/Package.swift
@@ -266,6 +266,7 @@ var targets: [PackageDescription.Target] = [
             .product(name: "SwiftASN1", package: "swift-asn1"),
             .product(name: "ServiceLifecycle", package: "swift-service-lifecycle"),
             .product(name: "AsyncAlgorithms", package: "swift-async-algorithms"),
+            .product(name: "Logging", package: "swift-log"),
         ],
         swiftSettings: strictConcurrencySettings
     ),
@@ -307,6 +308,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-asn1.git", from: "1.3.1"),
         .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.8.0"),
         .package(url: "https://github.com/apple/swift-async-algorithms.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-log.git", from: "1.6.3"),
 
     ],
     targets: targets
--- a/Package.swift
+++ b/Package.swift
@@ -263,7 +263,8 @@ var targets: [PackageDescription.Target] = [
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "X509", package: "swift-certificates"),
-            .product(name: "SwiftASN1", package: "swift-asn1")
+            .product(name: "SwiftASN1", package: "swift-asn1"),
+            .product(name: "ServiceLifecycle", package: "swift-service-lifecycle"),
         ],
         swiftSettings: strictConcurrencySettings
     ),
@@ -303,6 +304,8 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-certificates.git", branch: "1.10.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.29.3"),
         .package(url: "https://github.com/apple/swift-asn1.git", from: "1.3.1"),
+        .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.8.0"),
+
     ],
     targets: targets
 )
--- a/Package.swift
+++ b/Package.swift
@@ -265,6 +265,7 @@ var targets: [PackageDescription.Target] = [
             .product(name: "X509", package: "swift-certificates"),
             .product(name: "SwiftASN1", package: "swift-asn1"),
             .product(name: "ServiceLifecycle", package: "swift-service-lifecycle"),
+            .product(name: "AsyncAlgorithms", package: "swift-async-algorithms"),
         ],
         swiftSettings: strictConcurrencySettings
     ),
@@ -305,6 +306,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.29.3"),
         .package(url: "https://github.com/apple/swift-asn1.git", from: "1.3.1"),
         .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.8.0"),
+        .package(url: "https://github.com/apple/swift-async-algorithms.git", from: "1.0.0"),
 
     ],
     targets: targets
--- a/Package.swift
+++ b/Package.swift
@@ -275,7 +275,7 @@ var targets: [PackageDescription.Target] = [
             .product(name: "NIOCore", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "X509", package: "swift-certificates"),
-            .product(name: "SwiftASN1", package: "swift-asn1")
+            .product(name: "SwiftASN1", package: "swift-asn1"),
         ],
         swiftSettings: strictConcurrencySettings
     ),
--- a/Package.swift
+++ b/Package.swift
@@ -257,6 +257,27 @@ var targets: [PackageDescription.Target] = [
         ],
         swiftSettings: strictConcurrencySettings
     ),
+    .target(
+        name: "NIOCertificateHotReloading",
+        dependencies: [
+            .product(name: "NIOCore", package: "swift-nio"),
+            .product(name: "NIOSSL", package: "swift-nio-ssl"),
+            .product(name: "X509", package: "swift-certificates"),
+            .product(name: "SwiftASN1", package: "swift-asn1")
+        ],
+        swiftSettings: strictConcurrencySettings
+    ),
+    .testTarget(
+        name: "NIOCertificateHotReloadingTests",
+        dependencies: [
+            "NIOCertificateHotReloading",
+            .product(name: "NIOCore", package: "swift-nio"),
+            .product(name: "NIOSSL", package: "swift-nio-ssl"),
+            .product(name: "X509", package: "swift-certificates"),
+            .product(name: "SwiftASN1", package: "swift-asn1")
+        ],
+        swiftSettings: strictConcurrencySettings
+    ),
 ]
 
 let package = Package(
@@ -270,6 +291,7 @@ let package = Package(
         .library(name: "NIOHTTPTypesHTTP2", targets: ["NIOHTTPTypesHTTP2"]),
         .library(name: "NIOResumableUpload", targets: ["NIOResumableUpload"]),
         .library(name: "NIOHTTPResponsiveness", targets: ["NIOHTTPResponsiveness"]),
+        .library(name: "NIOCertificateHotReloading", targets: ["NIOCertificateHotReloading"]),
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.81.0"),
@@ -278,7 +300,9 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-http-structured-headers.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-algorithms.git", from: "1.2.0"),
-
+        .package(url: "https://github.com/apple/swift-certificates.git", branch: "1.10.0"),
+        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.29.3"),
+        .package(url: "https://github.com/apple/swift-asn1.git", from: "1.3.1"),
     ],
     targets: targets
 )
--- a/Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
+++ b/Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
@@ -48,8 +48,8 @@ public protocol CertificateReloader: Sendable {
 /// Once the reloader is running, call ``sslContextConfigurationOverride`` to get a
 /// `NIOSSLContextConfigurationOverride` which can be set on NIO's `TLSConfiguration`: this will keep the certificate
 /// and private key pair up to date.
-/// You may instead call `TLSConfiguration.withAutomaticCertificateReloading(using:)` to get a
-/// `TLSConfiguration` with a configured reloader.
+/// You may instead call  ``NIOSSL/TLSConfiguration/setCertificateReloader(_:)`` to get a
+/// ``NIOSSL/TLSConfiguration`` with a configured reloader.
 ///
 /// If any errors occur during a reload attempt (such as: being unable to find the file(s) containing the certificate or the key; the format
 /// not being recognized or not matching the configured one; not being able to verify a certificate's signature against the given
--- a/Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
+++ b/Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
@@ -12,11 +12,13 @@
 //
 //===----------------------------------------------------------------------===//
 
-import X509
-import NIOSSL
-import struct NIOCore.TimeAmount
 import NIOConcurrencyHelpers
+import NIOSSL
 import ServiceLifecycle
+import X509
+
+import struct NIOCore.TimeAmount
+
 #if canImport(FoundationEssentials)
 import FoundationEssentials
 #else
@@ -130,7 +132,6 @@ public struct TimedCertificateReloader: CertificateReloader {
     private let privateKeyDescription: PrivateKeyDescription
     private let state: NIOLockedValueBox<CertificateKeyPair?>
 
-
     /// A `NIOSSLContextConfigurationOverride` that will be used as part of the NIO application's TLS configuration.
     /// Its certificate and private key will be kept up-to-date via the reload mechanism the ``TimedCertificateReloader``
     /// implementation provides.
@@ -165,7 +166,7 @@ public struct TimedCertificateReloader: CertificateReloader {
         // reload loop to run.
         self.reloadPair()
     }
-    
+
     /// A long-running method to run the ``TimedCertificateReloader`` and start observing updates for the certificate and
     /// private key pair.
     /// - Important: You *must* call this method to get certificate and key updates.
@@ -178,10 +179,11 @@ public struct TimedCertificateReloader: CertificateReloader {
 
     private func reloadPair() {
         if let certificateBytes = self.loadCertificate(),
-           let keyBytes = self.loadPrivateKey(),
-           let certificate = self.parseCertificate(from: certificateBytes),
-           let key = self.parsePrivateKey(from: keyBytes),
-           key.publicKey.isValidSignature(certificate.signature, for: certificate) {
+            let keyBytes = self.loadPrivateKey(),
+            let certificate = self.parseCertificate(from: certificateBytes),
+            let key = self.parsePrivateKey(from: keyBytes),
+            key.publicKey.isValidSignature(certificate.signature, for: certificate)
+        {
             self.attemptToUpdatePair(certificate: certificate, key: key)
         }
     }
@@ -189,7 +191,7 @@ public struct TimedCertificateReloader: CertificateReloader {
     private func loadCertificate() -> [UInt8]? {
         let certificateBytes: [UInt8]?
         switch self.certificateDescription.location._backing {
-        case .file(path: let path):
+        case .file(let path):
             let bytes = FileManager.default.contents(atPath: path)
             certificateBytes = bytes.map { Array($0) }
 
@@ -202,7 +204,7 @@ public struct TimedCertificateReloader: CertificateReloader {
     private func loadPrivateKey() -> [UInt8]? {
         let keyBytes: [UInt8]?
         switch self.privateKeyDescription.location._backing {
-        case .file(path: let path):
+        case .file(let path):
             let bytes = FileManager.default.contents(atPath: path)
             keyBytes = bytes.map { Array($0) }
 
--- a/Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
+++ b/Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
@@ -16,6 +16,7 @@ import X509
 import NIOSSL
 import struct NIOCore.TimeAmount
 import NIOConcurrencyHelpers
+import ServiceLifecycle
 #if canImport(FoundationEssentials)
 import FoundationEssentials
 #else
@@ -271,3 +272,6 @@ extension TLSConfiguration {
         return self
     }
 }
+
+@available(macOS 11.0, iOS 14, tvOS 14, watchOS 7, *)
+extension TimedCertificateReloader: Service {}
--- a/Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
+++ b/Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
@@ -158,9 +158,11 @@ public struct TimedCertificateReloader: CertificateReloader {
         self.refreshInterval = refreshInterval
         self.certificateDescription = certificateDescription
         self.privateKeyDescription = privateKeyDescription
-
-        // TODO: try parsing key and cert here too
         self.state = NIOLockedValueBox(nil)
+
+        // Immediately try to load the configured cert and key to avoid having to wait for the first
+        // reload loop to run.
+        self.reloadPair()
     }
     
     /// A long-running method to run the ``TimedCertificateReloader`` and start observing updates for the certificate and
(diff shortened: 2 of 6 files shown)


Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,279 @@
...
+
+        /// This certificate/key can be found at the given filepath.
+        /// - Parameter path: The filepath where the certificate/key can be found.
+        /// - Returns: A `Location`.
+        public static func file(path: String) -> Self { Self(_Backing.file(path: path)) }
+
+        /// This certificate/key is available in memory, and will be provided by the given closure.
+        /// - Parameter provider: A closure providing the bytes for the given certificate or key.
+        /// - Returns: A `Location`.
+        public static func memory(provider: @Sendable @escaping () -> [UInt8]?) -> Self {
```
FranzBusch, line 98: Can we document in which case this closure should return `nil`?

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,279 @@
...
+            Self(_Backing.memory(provider: provider))
+        }
+    }
+
+    /// A description of a certificate, in terms of its ``Location`` and ``Encoding``.
+    public struct CertificateDescription: Sendable {
+        public var location: Location
+        public var format: Encoding
```
FranzBusch, line 106: NIT docs and in a few more public APIs

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+    /// Initialize a new ``TimedCertificateReloader``.
+    /// - Parameters:
+    ///   - refreshInterval: The interval at which attempts to update the certificate and private key should be made.
+    ///   - certificateDescription: A ``CertificateDescription``.
+    ///   - privateKeyDescription: A ``PrivateKeyDescription``.
+    public init(
+        refreshingEvery refreshInterval: TimeAmount,
```
FranzBusch, line 156: Why are we not taking a `Duration` here?

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,279 @@
...
+            )
+        }
+    }
+}
+
+extension TLSConfiguration {
+    /// Configure a ``CertificateReloader`` to observe updates for the certificate and key pair used.
+    /// - Parameter reloader: A ``CertificateReloader`` to watch for certificate and key pair updates.
+    /// - Returns: A `TLSConfiguration` that reloads the certificate and key used in its SSL handshake.
+    mutating public func withAutomaticCertificateReloading(using reloader: any CertificateReloader) -> Self {
```
FranzBusch, line 270: Why are we using an existential here instead of `some`?

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,279 @@
...
+///
+/// A certificate reloader is a service that can provide you with updated versions of a certificate and private key pair, in
+/// the form of a `NIOSSLContextConfigurationOverride`, which will be used when performing a TLS handshake in NIO.
+/// Each implementation can choose how to observe for changes, but they all require an ``sslContextConfigurationOverride``
+/// to be exposed.
+public protocol CertificateReloader: Sendable {
+    /// A `NIOSSLContextConfigurationOverride` that will be used as part of the NIO application's TLS configuration.
+    /// Its certificate and private key will be kept up-to-date via whatever mechanism the specific ``CertificateReloader``
+    /// implementation provides.
+    var sslContextConfigurationOverride: NIOSSLContextConfigurationOverride { get }
```
FranzBusch, line 38: Do we ever foresee a reloader that wants to load the configuration `async`?

FranzBusch (commented):
This is exciting! Making TLS configuration updates easier is something that has been requested a lot

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: Because it's only available on macOS 13+. Using `TimeAmount` we can support 11+. I was on the fence about this.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: Oversight :)

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: I can't think of an obvious example but yes in theory I don't see why not. I think for extensibility we should probably make this `async` in the protocol.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,301 @@
...
+/// A certificate reloader is a service that can provide you with updated versions of a certificate and private key pair, in
+/// the form of a `NIOSSLContextConfigurationOverride`, which will be used when performing a TLS handshake in NIO.
+/// Each implementation can choose how to observe for changes, but they all require an ``sslContextConfigurationOverride``
+/// to be exposed.
+@available(macOS 11.0, iOS 14, tvOS 14, watchOS 7, *)
+public protocol CertificateReloader: Sendable {
+    /// A `NIOSSLContextConfigurationOverride` that will be used as part of the NIO application's TLS configuration.
+    /// Its certificate and private key will be kept up-to-date via whatever mechanism the specific ``CertificateReloader``
+    /// implementation provides.
+    var sslContextConfigurationOverride: NIOSSLContextConfigurationOverride { get async }
```
FranzBusch, line 40: I have been thinking about this some more. There is two ways to go with this here. Keep it close to NIO then it should probably not be async but rather returning an ELF. Or this is a protocol that we see as a bridge to the Swift Concurrency ecosystem then we should make it async but I also think we should be returning a type here that is based on swift-certificates and map to NIOSSL types under the hood.

I think in general this should become a method though and not a computed property.

Would be interested to hear @Lukasa thoughts here

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
FranzBusch: I see. Can we add a second init with duration then that has an availability annotation?

gjcairo: Yeah good callout.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
glbrntt: Isn't the intention for this to effectively be a current value? The reloader implementation does any work it actually needs be it async or otherwise.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,301 @@
...
+        }
+    }
+}
+
+extension TLSConfiguration {
+    /// Configure a ``CertificateReloader`` to observe updates for the certificate and key pair used.
+    /// - Parameter reloader: A ``CertificateReloader`` to watch for certificate and key pair updates.
+    /// - Returns: A `TLSConfiguration` that reloads the certificate and key used in its SSL handshake.
+    @available(macOS 11.0, iOS 14, tvOS 14, watchOS 7, *)
+    mutating public func withAutomaticCertificateReloading(using reloader: some CertificateReloader) -> Self {
```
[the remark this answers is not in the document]
gjcairo: Fair. What about `enableAutomaticCertificateReloading`?

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,319 @@
...
+        }
+    }
+
+    /// Initialize a new ``TimedCertificateReloader``.
+    /// - Parameters:
+    ///   - refreshInterval: The interval at which attempts to update the certificate and private key should be made.
+    ///   - certificateDescription: A ``CertificateDescription``.
+    ///   - privateKeyDescription: A ``PrivateKeyDescription``.
+    public init(
+        refreshingEvery refreshInterval: TimeAmount,
```
glbrntt, line 177: nit: from the [API design guidelines](https://www.swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage):

> The first argument to initializer and [factory methods](https://en.wikipedia.org/wiki/Factory_method_pattern) calls should not form a phrase starting with the base name

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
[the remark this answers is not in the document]
glbrntt: IMO `enable` suggests a boolean e.g. `enableAutomaticCertificateReloading = true`.

My bias is normally towards simple: `setCertificateReloader(_:)`.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /// Initialize a new ``TimedCertificateReloader``.
+    /// - Parameters:
+    ///   - refreshInterval: The interval at which attempts to update the certificate and private key should be made.
+    ///   - certificateDescription: A ``CertificateDescription``.
+    ///   - privateKeyDescription: A ``PrivateKeyDescription``.
+    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
+    public init(
+        refreshingEvery refreshInterval: Duration,
```
glbrntt, line 198: Do we run the risk of having ambiguity issues with the `Duration`/`TimeAmount` overload? Both have similar static funcs for e.g. `.seconds` etc.

gjcairo: I don't actually know the answer to this but at least on the tests (even when adding an availability annotation to match the `Duration` overload) it resolves to using `TimeAmount`. I am unsure if some other combination of factors may cause the compiler to become confused.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,319 @@
...
+
+/// A ``TimedCertificateReloader`` is an implementation of a ``CertificateReloader``, where the certificate and private
+/// key pair is updated at a fixed interval from the file path or memory location configured.
+///
+/// You initialize a ``TimedCertificateReloader`` by providing a refresh interval, and locations for the certificate and the private
+/// key. You must then call ``run()`` on this reloader for it to start observing changes.
+/// Once the reloader is running, call ``sslContextConfigurationOverride`` to get a
+/// `NIOSSLContextConfigurationOverride` which can be set on NIO's `TLSConfiguration`: this will keep the certificate
+/// and private key pair up to date.
+/// You may instead call `TLSConfiguration.withAutomaticCertificateReloading(using:)` to get a
```
hamzahrmalik, line 51: should this be setCertificateReloader? And should it be in `` so it becomes a hyperlink?

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,319 @@
...
+}
+
+extension TLSConfiguration {
+    /// Configure a ``CertificateReloader`` to observe updates for the certificate and key pair used.
+    /// - Parameter reloader: A ``CertificateReloader`` to watch for certificate and key pair updates.
+    /// - Returns: A `TLSConfiguration` that reloads the certificate and key used in its SSL handshake.
+    @available(macOS 11.0, iOS 14, tvOS 14, watchOS 7, *)
+    mutating public func setCertificateReloader(_ reloader: some CertificateReloader) -> Self {
+        self.sslContextCallback = { _, promise in
+            promise.completeWithTask { await reloader.sslContextConfigurationOverride }
```
hamzahrmalik, line 312: is there any way we could avoid this unstructured task? Perhaps by having some run() function load the latest thing out continually into a a mutex or something which can be read from here?

(I havent looked at the rest of the code yet, so not sure if that makes sense, but ive hit weird issues with promise.completeWithTask before)

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: Yeah good catch

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,319 @@
...
+            refreshInterval: TimeAmount(refreshInterval),
+            certificateDescription: certificateDescription,
+            privateKeyDescription: privateKeyDescription
+        )
+    }
+
+    /// A long-running method to run the ``TimedCertificateReloader`` and start observing updates for the certificate and
+    /// private key pair.
+    /// - Important: You *must* call this method to get certificate and key updates.
+    public func run() async throws {
```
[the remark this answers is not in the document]
gjcairo: It is yeah - we want to make sure we respect cancellation. The interval may be set to a pretty high value (you likely only want to reload your certs every e.g. 10 minutes) so cancellation is important.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: Well, if we keep the `sslContextConfigurationOverride` property `async` in the `CertificateReloader` protocol I don't think there's a way around this. There's some discussion above about whether we actually want this to be async or not.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: In theory someone may want to avoid having a long running `run` for some reason?
Or another possible scenario would be if you were to yield new pairs into an async stream continuation from the `run` method, and iterate the stream in this getter instead of updating some [protected] state (like a LockedValueBox) containing the new values.

In any case, I do think that this is all avoidable and you can handle the asynchrony from elsewhere (like a `run` method). But not sure if Franz had something else in mind. I suppose having it `async` gives implementers some more freedom.

Package.swift
```diff
@@ -257,6 +257,28 @@ var targets: [PackageDescription.Target] = [
         ],
         swiftSettings: strictConcurrencySettings
     ),
+    .target(
+        name: "NIOCertificateHotReloading",
```
czechboy0, line 261: NIOCertificateReloading might be enough?

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,319 @@
...
+        while !Task.isCancelled {
```
czechboy0, line 213: Might be better to do `while true` to keep behavior consistent regardless of when exactly cancelation happens. Otherwise sometimes you might end up with a clean return and with a thrown cancelation error otherwise.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,319 @@
...
+            return override
+        }
+    }
+
+    /// Initialize a new ``TimedCertificateReloader``.
+    /// - Parameters:
+    ///   - refreshInterval: The interval at which attempts to update the certificate and private key should be made.
+    ///   - certificateDescription: A ``CertificateDescription``.
+    ///   - privateKeyDescription: A ``PrivateKeyDescription``.
+    public init(
```
czechboy0, line 176: Could we get an initializer that throws an error if it fails the first reload? Otherwise it might be harder to debug eg an incorrect path to the cert/key.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
czechboy0: I'd prefer if it's not async. Otherwise calling it from synchronous code requires more complicated state management, etc.

czechboy0: I agree with George's assessment - my understanding was that this represents the "current" value. If so, I think it should be non-throwing, and sync.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
czechboy0: In other words - failing the first load is fatal, and the server probably wants to fail to start. Failing to reload later can be handled more gracefully - just keep the old value.

It'd be great to make that distinction easier to react to in a server startup sequence. One way is to represent this as an async throwing initializer that performs (or fails) the first load.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
```diff
@@ -0,0 +1,319 @@
...
+            keyBytes = bytesProvider()
+        }
+        return keyBytes
+    }
+
+    private func parseCertificate(from certificateBytes: [UInt8]) -> Certificate? {
+        let certificate: Certificate?
+        switch self.certificateDescription.format._backing {
+        case .der:
+            certificate = try? Certificate(derEncoded: certificateBytes)
```
czechboy0, line 260: We should let the error be thrown out and at least logged (here and all the `try?` below as well)

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
glbrntt: Good point. Makes me wonder whether this should throw at all because with that change it can never return cleanly. (Not sure this actually matters much.)

czechboy0: Yeah it'd be good to get more guidance on the expected behavior - should services return or throw CancellationError. I've seen us do both.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: It _is_ meant to represent the current state yes. After this discussion and some more thought I'm also thinking I should revert this to be sync.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: Yeah good callout - I'll switch to `while true`.

With regards to whether it should be throwing or we should return, a (maybe useful?) data point, `ServiceLifecycle` will throw `CancellationError` if you try gracefully shutting down a service that's already been cancelled - even if the service itself didn't throw `CancellationError` when it was cancelled. Also, many of the async APIs Apple provides (like `Task.sleep` in this case, or async iterators' `next`) will handle cancellation by throwing a `CancellationError` too. 

I think throwing out of this `Task.sleep` is particularly useful because you don't want to wait for the whole timeout period, which could be minutes, to finish before the reloading task is cancelled. I do think it's the right thing to do here.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: > Could we get an initializer that throws an error if it fails the first reload? Otherwise it might be harder to debug eg an incorrect path to the cert/key.

I originally flip flopped a bit on this. I ended up landing on not failing because what if your certificate is not yet present but will be? You could have a daemon that takes care of managing your TLS certificates and this could be the first time you're starting your service on a machine and the cert is not yet present, or the cert could be missing for whatever other reason (data corruption, mistakenly deleted, etc). 

I was trying to find what other projects do in this scenario in case we want to get some inspiration. `cert-manager` ([link](https://github.com/cert-manager/cert-manager/blob/f866efad30e60a1ad1d9f4ec9e7ffbbc29245d5e/pkg/server/tls/file_source.go#L83)) allows for this possibility and does not fail on first load.

However it isn't lost on me that not failing could potentially make debugging harder. It could also mean an insecure connection might be established with a client which obviously isn't good. So I'm okay with making this init throwing. I just want to make sure we've considered both options.

czechboy0: Yeah good on you for being thoughtful on this, and I'm not even saying there shouldn't be a graceful initializer that ignores the first failure as well. I mainly just wanted to have the option to have a throwing initializer.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: I am not sure about throwing - I don't think we want the reloader to stop running when some potentially transient error is encountered. 

We could have a max number of failed attempts though, and throw after that number has been reached.

W.r.t logging, we don't currently depend on `swift-log` in this package, so I'm not sure if that's something we want. I guess we could also just `print` (which I know is not ideal but it's what we already do for the [debug events handler](https://github.com/apple/swift-nio-extras/blob/f1f6f772198bee35d99dd145f1513d8581a54f2c/Sources/NIOExtras/DebugInboundEventsHandler.swift#L155))

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: I think it's reasonable to offer this, I'll add additional inits that validate they can be loaded.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: Okay so given the other discussion about having throwing inits, I _have_ made these methods throw, but I am just ignoring the errors from the main reload loop.

Still open questions about having a max number of failed attempts and a logger.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: I've reverted the property to be sync so the unstructured task's gone.

hamzahrmalik:
What's the best way to use this with a server?

```swift
let tlsConfig = TLSConfiguration.makeServerConfiguration(certificateChain: ??????) // What do i pass here?
tlsConfig.sslContextCallback = { _, promise in
    promise.succeed(timedCertificateReloader.sslContextConfigurationOverride)
}
```
It'd be nice if TimedCertificateReloader could expose the initial config, since it already read it in its initializer, or if TLSConfiguration could be made without an initial cert-key pair (and rely on the override).

Or do i have to just manually read the first cert-key pair?

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
hamzahrmalik: im somewhat suspicious that there's a way to break this if you write

init(refreshInterval: .seconds(1)...
the compiler won't know if you meant Duration.seconds or TimeAmount.seconds

That said, i can't come up with a way to make this break

hamzahrmalik: im thinking about https://github.com/apple/swift-metrics/pull/144

Sources/NIOCertificateReloading/CertificateReloader.swift
```diff
@@ -0,0 +1,84 @@
...
+            certificateChain: certificateChain,
+            privateKey: privateKey
+        )
+        return configuration.setCertificateReloader(certificateReloader)
+    }
+
+    /// Configure a ``CertificateReloader`` to observe updates for the certificate and key pair used.
+    /// - Parameter reloader: A ``CertificateReloader`` to watch for certificate and key pair updates.
+    /// - Returns: A ``NIOSSL/TLSConfiguration`` that reloads the certificate and key used in its SSL handshake.
+    mutating public func setCertificateReloader(_ reloader: some CertificateReloader) -> Self {
```
hamzahrmalik, line 78: this is mildly confusing to me, it's mutating but also returns self

If the return is just for convenience, then we should maybe make it discardable result

Or was it intended to not mutate the original and make a copy?

hamzahrmalik: also, xcode says you need to import NIOCore
```
Instance method 'succeed' is not available due to missing import of defining module 'NIOCore'
```

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -0,0 +1,400 @@
...
+        try self.reloadPair()
+    }
+
+    /// A long-running method to run the ``TimedCertificateReloader`` and start observing updates for the certificate and
+    /// private key pair.
+    /// - Important: You *must* call this method to get certificate and key updates.
+    public func run() async throws {
+        for try await _ in AsyncTimerSequence.repeating(every: self.refreshInterval).cancelOnGracefulShutdown() {
+            // We don't want to throw out of this method so simply ignore errors thrown from `reloadPair`.
+            try? self.reloadPair()
```
czechboy0, line 310: Could benefit from some logging integration, at least a callback when reload fails? I'd want to hook it into logging/metrics when running in a service.

Sources/NIOCertificateReloading/CertificateReloader.swift
gjcairo: It's for convenience, so you can chain calls. I've added `@discardableResult`.

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
gjcairo: Added logging when the reloading fails.

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -0,0 +1,421 @@
...
+        refreshInterval: TimeAmount,
+        validatingCertificateDescription: CertificateDescription,
+        validatingPrivateKeyDescription: PrivateKeyDescription,
+        logger: Logger? = nil
+    ) throws {
+        try self.init(
+            refreshInterval: Duration(refreshInterval),
+            validatingCertificateDescription: validatingCertificateDescription,
+            validatingPrivateKeyDescription: validatingPrivateKeyDescription,
+            logger: nil
```
czechboy0, line 262:
```suggestion
            logger: logger
```

Or was it intentional?

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -0,0 +1,421 @@
...
+
+    /// A long-running method to run the ``TimedCertificateReloader`` and start observing updates for the certificate and
+    /// private key pair.
+    /// - Important: You *must* call this method to get certificate and key updates.
+    public func run() async throws {
+        for try await _ in AsyncTimerSequence.repeating(every: self.refreshInterval).cancelOnGracefulShutdown() {
+            do {
+                try self.reloadPair()
+            } catch {
+                self.logger?.error(
```
czechboy0, line 322:
```suggestion
                self.logger?.debug(
```

See https://www.swift.org/documentation/server/guides/libraries/log-levels.html

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -0,0 +1,421 @@
...
+                    """
+                    An unexpected error was encountered while trying to reload the certificate and \
+                    private key pair.
+                    """,
+                    metadata: [
```
czechboy0, line 327: Please also attach the paths to the files that it was trying to load.

czechboy0 (commented):
Thanks, Gus!

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -0,0 +1,421 @@
...
+    ) {
+        self.init(
+            refreshInterval: Duration(refreshInterval),
+            certificateDescription: certificateDescription,
+            privateKeyDescription: privateKeyDescription,
+            logger: logger
+        )
+    }
+
+    /// Attempt to initialize a new ``TimedCertificateReloader``, but throw if the given certificate and private keys cannot be
```
glbrntt, line 245: Can I propose that rather than having `init`s which also do the first refresh and can throw that we just have the non-throwing `init`s and also expose a `refreshNow()` throwing func?

This means that:
- the `init` doesn't have side effects
- refreshing is explicit
- any throwing is directly tied to the refresh operation (throwing _could_ come from something else in the `init` but as a user I don't know that unless I read the code)
- we can remove some duplication

Generally speaking, if work should be done after creating the object it should either be a separate func which can be made more convenient by a `static func` which creates and then runs that func before returning the object.

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -0,0 +1,421 @@
...
+    /// - Important: You *must* call this method to get certificate and key updates.
+    public func run() async throws {
+        for try await _ in AsyncTimerSequence.repeating(every: self.refreshInterval).cancelOnGracefulShutdown() {
+            do {
+                try self.reloadPair()
+            } catch {
+                self.logger?.error(
+                    """
+                    An unexpected error was encountered while trying to reload the certificate and \
+                    private key pair.
```
glbrntt, line 325: General rule-of-thumb, aim to be concise. If you can say it as clearly in fewer words then you should: "Failed to reload certificate and private key"

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -0,0 +1,421 @@
...
+                    """,
+                    metadata: [
+                        "error": error
```
glbrntt, line 328: Can you add tests for the description of the errors that can be thrown? Structs wrapping enums have bad default descriptions IIRC.

Tests/NIOCertificateReloadingTests/TimedCertificateReloaderTests.swift
```diff
@@ -0,0 +1,322 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@preconcurrency import Crypto
```
glbrntt, line 15: What's used that requires `@preconcurrency`?

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
gjcairo: Nope, copy pasta

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
gjcairo: So are you suggesting that on top of the `reloadNow()` function, we also provide a static `TimedCertificateReloader.makeReloaderAndValidateLocations(...)` (naming TBC)?

@czechboy0 - what do you think of this change? Tagging you since you originally asked for throwing inits.

Personally, if we also provide a static method that does both things, I'm okay with it.

glbrntt: I (personally) wouldn't bother adding the `static` method as well but I have no objections to it being added.

czechboy0: I'm fine with either a throwing init or a throwing static func (they're mostly equivalent in my mind), I just want it to be a single call that creates and returns the instance, and we know the instance was able to load the certs at least once (meaning we don't need code that handles one of these that never successfully loaded certs).

I suspect we'll add that call to many of our examples and docs, as I believe that a service that can't load certs at least once shouldn't ever mark itself healthy. But I'm okay with that not being the only usage pattern here, in case there are use cases where someone wants to keep running even without certs ever loading successfully.

gjcairo: Not being able to load the certs on the first try doesn't mean they won't ever load successfully :) but I agree this could be a common pattern so I do think we should offer a way to do so easily. 

Because of the switch from `while true` in the main loop to using the timer sequence, I've had to increase the platforms in the availability guards anyways. So I'll just leave the non-throwing `init` using `Duration` (and get rid of the `TimeAmount` overload, since it was only there to support older platforms which can't be supported anymore), add a new `reloadNow()` func, and add a new static method to init and reload.

gjcairo: I don't personally see having to use a static method instead of an init as a more complicated alternative. 

But to answer your question, I wouldn't expect a server to run long-term without the cert and private key pair in place. My idea of why tolerating this on the server would be useful is that if the initial provisioning races with the service's startup and takes slightly longer, then we won't just fail the whole service group but rather recover soon after - even if in the meantime the server chooses to e.g. refuse connections. And we _are_ logging failed reload attempts so we wouldn't silently downgrade to HTTP if you have observability rightly set up.

Also: this can be used on the client-side too, it isn't just for servers. You may be using mTLS, where you also need to rotate your certs. And in that case, you may want to still create a client but e.g. queue requests until a valid cert is available.

Sources/NIOCertificateHotReloading/NIOCertificateHotReloading.swift
gjcairo: Because I had to up the availability requirements to use the async timer sequence, we can just leave the `Duration` init - I got rid of the `TimeAmount` overload.

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
czechboy0: I see, I think we're each looking at it from the perspective of different users.

I'm thinking of the operators of a full service. Those users still use the TLSConfiguration type directly, as most (maybe all major) server frameworks don't wrap it, but instead directly take it as a parameter. For those users, I think it could be an issue that the "simple" way to create the cert rotater is actually the unsafe one, and they likely don't have deep familiarity with what the server framework does in cases of no certs (I use Vapor, but don't actually know, without checking, how it would handle this case once it adopts this API).

Correct me if I'm wrong, you're thinking of the implementors of other libraries on top of this package, who do make these decisions (what happens if we don't successfully load a cert in the first 10 minutes, for example - what does the server do?)

Now I wonder if these just shouldn't be two separate types, and each library can adopt the one it actually needs. One that is guaranteed to never have empty certs, and another one that allows it. And as a server library author, I'd never use the second one, because it's a new failure mode that my library now needs to handle, but didn't have to handle when you take immutable, non-optional certs an input.

I agree that server bootstrap can be racy, but I haven't seen that to be the case for certificates specifically. Loading certs from disk is not expected to be flaky under normal conditions, unlike, for example, an upstream HTTP service dependency (which has a <100% SLO, so we need to design for that).

Are we assuming here loading from disk, or the underlying platform lying down the certs, are also expected to be flaky to the point of us wanting to handle it gracefully? My lean is: no, but if this is a requirement for some users, I'm ready to believe this package should support it. I just think it should be more explicit in the naming, and the default behavior should be the safe, conservative one, rather than the potentially unsafe, more niche one.

But that's just my 2c, it's possible I haven't considered some common use case or requirement 🙂

glbrntt: Just making up APIs here, but hopefully it's clear enough:

```swift
let reloader = CertificateReloader(refreshInterval: .minutes(60))
let server = SomeServer(tls: .makeServerConfig(using: reloader))
let serviceGroup = ServiceGroup(services: [reloader, server])
try await serviceGroup.run()
```

If `reloader` throws on `run()` (which should only be possible when doing the _first_ refresh) then the service group will get shutdown almost immediately. I _think_ there's a small timing window between the reloader starting and failing where the server _could_ start but would be a lame duck. In practice this is unlikely to be an issue as the server should be failing its health checks and the timing window before shutting down should be very small.

Sources/NIOCertificateReloading/CertificateReloader.swift
```diff
@@ -0,0 +1,97 @@
...
+
+        /// The given ``CertificateReloader`` could not provide a private key with which to create this config.
+        public static var missingPrivateKey: Self { .init(backing: .missingPrivateKey) }
+    }
+
+    /// Create a ``NIOSSL/TLSConfiguration`` for use with server-side contexts, with certificate reloading enabled.
+    /// - Parameter certificateReloader: A ``CertificateReloader`` to watch for certificate and key pair updates.
+    /// - Returns: A ``NIOSSL/TLSConfiguration`` for use with server-side contexts, that reloads the certificate and key
+    /// used in its SSL handshake.
+    public static func makeServerConfiguration(
```
czechboy0, line 69: Could we also get a makeClientConfiguration variant for mTLS use cases?

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
czechboy0: Is this meant to be the same `makeServerConfiguration` as in this PR? Because that's a throwing function that will throw an error if the cert hasn't been loaded yet, so it can only be practically used with CertificateReloader that is guaranteed to have already loaded certs (my preference).

glbrntt: Ah right, yeah I misremembered the API. Let's just go with the static, it's easy to understand.

czechboy0: Cool, if we also get the makeClient variant, I'm happy with this PR.

czechboy0 (approved):
Looks great, thank you @gjcairo 🙏

Sources/NIOCertificateReloading/CertificateReloader.swift
```diff
@@ -0,0 +1,123 @@
...
+    /// - Returns: A ``NIOSSL/TLSConfiguration`` for use with client-side contexts, that reloads the certificate and key
+    /// used in its SSL handshake.
+    /// - Throws: This method will throw if an override isn't present. This may happen if a certificate or private key could not be
+    /// loaded from the given paths.
+    public static func makeClientConfiguration(
+        certificateReloader: some CertificateReloader
+    ) throws -> Self {
+        let override = certificateReloader.sslContextConfigurationOverride
+
+        guard override.certificateChain != nil else {
```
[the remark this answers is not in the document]
gjcairo: Hm, I'm actually not sure it's needed since the reloader takes care of setting the override via `sslContextCallback`, but I'll set it just in case.

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -0,0 +1,440 @@
...
+/// ``TimedCertificateReloader`` conforms to `ServiceLifecycle`'s `Service` protocol, meaning you can simply create
+/// the reloader and add it to your `ServiceGroup` without having to manually run it.
+///
+/// If any errors occur during a reload attempt (such as: being unable to find the file(s) containing the certificate or the key; the format
+/// not being recognized or not matching the configured one; not being able to verify a certificate's signature against the given
+/// private key; etc), then that attempt will be aborted but the service will keep on trying at the configured interval.
+/// The last-valid certificate-key pair (if any) will be returned as the ``sslContextConfigurationOverride``.
+#if compiler(>=6.0)
+@available(macOS 13, iOS 16, watchOS 9, tvOS 16, macCatalyst 16, visionOS 1, *)
+#else
```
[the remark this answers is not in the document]
gjcairo: Good point - I think that should be fine yes. Do we care about explicitly spelling out all platforms or not really? I know we've been adding visionOS to other packages, because we weren't using `*`, but should we opted to use `*` instead? Or do we favour listing them explicitly?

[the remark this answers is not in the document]
glbrntt: I don't think so. I'm not opposed to spelling it out fully, it just seems overkill to compiler guard it (especially if it was applied more widely)

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -0,0 +1,440 @@
...
+        }
+
+        /// This certificate/key can be found at the given filepath.
+        /// - Parameter path: The filepath where the certificate/key can be found.
+        /// - Returns: A `Location`.
+        public static func file(path: String) -> Self { Self(_Backing.file(path: path)) }
+
+        /// This certificate/key is available in memory, and will be provided by the given closure.
+        /// - Parameter provider: A closure providing the bytes for the given certificate or key. This closure should return
+        /// `nil` if a certificate/key isn't currently available for whatever reason.
```
glbrntt, line 161: Docs are wrong now re: `nil`

glbrntt (approved):
docs nit but LGTM otherwise

PR #273, opened 2025-05-20
Support certificate chains from single PEM in `TimedCertificateReloader`

FranzBusch:
## Motivation

The new `TimedCertificateReloader` currently only supports loading PEM files with a single cert. It is often the case that a single PEM file contains an entire certificate chain.

## Modification

Modifies the code to support loading an entire certificate chain from a single PEM file.

## Result

PEM based certificate chain support in our new `TimedCertificateReloader`

### DIFF
--- a/Tests/NIOCertificateReloadingTests/TimedCertificateReloaderTests.swift
+++ b/Tests/NIOCertificateReloadingTests/TimedCertificateReloaderTests.swift
@@ -31,7 +31,7 @@ final class TimedCertificateReloaderTests: XCTestCase {
         try await runTimedCertificateReloaderTest(
             certificate: .init(location: .file(path: "doesnotexist"), format: .der),
             privateKey: .init(
-                location: .memory(provider: { Array(Self.samplePrivateKey.derRepresentation) }),
+                location: .memory(provider: { Array(Self.samplePrivateKey1.derRepresentation) }),
                 format: .der
             ),
             validateSources: false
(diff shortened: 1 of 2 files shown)


Tests/NIOCertificateReloadingTests/TimedCertificateReloaderTests.swift
```diff
@@ -292,6 +292,52 @@ final class TimedCertificateReloaderTests: XCTestCase {
             )
         }
     }
+    
+    func testReloadSuccessfullyCertificateChain_FromMemory() async throws {
```
gjcairo, line 296: Would you mind adding a `_FromFile` equivalent?

Tests/NIOCertificateReloadingTests/TimedCertificateReloaderTests.swift
```diff
@@ -484,6 +530,42 @@ final class TimedCertificateReloaderTests: XCTestCase {
...
+                notValidAfter: startDate.advanced(by: 60 * 60 * 24 * 360),
+                issuer: sampleCertName,
+                subject: sampleCertName,
+                signatureAlgorithm: .ecdsaWithSHA384,
+                extensions: Certificate.Extensions {
+                    Critical(
+                        BasicConstraints.isCertificateAuthority(maxPathLength: nil)
+                    )
+                },
+                issuerPrivateKey: .init(samplePrivateKey)
```
gjcairo, line 565: Could you use a different private key here? I want to make sure we test that we're validating the key against the right cert (the first one).

gjcairo (changes requested):

gjcairo (approved):
Great, thanks Franz!

PR #278, opened 2025-06-18
Responsiveness fallback handler

josephnoir:
### Motivation:

The SimpleResponsivenessRequestMux knows how to handle a few different requests. For those it doesn’t know how to handle, it responds with a 404.

It would be helpful if the handler had a mode where it optionally forwards requests it doesn’t know how to handle to a fallback channel handler so that users can respond to these requests.

### Modifications:

The mux has a new initializer that can configure it to accept unexpected requests, i.e., those not defined by itself. In this case it will forward requests down the pipeline. However, its default behavior remains the same: respond with a 404.

### Result:

When setting up the mux, users can add additional handler to the pipeline and configure the mux to forward its requests.

### DIFF
--- a/Sources/NIOHTTPResponsiveness/SimpleResponsivenessRequestMux.swift
+++ b/Sources/NIOHTTPResponsiveness/SimpleResponsivenessRequestMux.swift
@@ -17,8 +17,17 @@ import HTTPTypes
 import NIOCore
 import NIOHTTPTypes
 
-/// Basic request multiplexer that identifies which request type (config, download, upload) is requested and adds the appropriate handler.
-/// Once the handler has been added, all data is passed through to the newly added handler.
+/// A basic request multiplexer that identifies which request type (config, download, upload) is requested and adds the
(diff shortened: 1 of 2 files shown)


Sources/NIOHTTPResponsiveness/SimpleResponsivenessRequestMux.swift
```diff
@@ -99,26 +135,24 @@ public final class SimpleResponsivenessRequestMux: ChannelInboundHandler {
...
+                }
             }
         }
 
-        // Only pass through data through if we've actually added a handler. If we didn't add a handler, it's because we
-        // directly responded. In this case, we don't care about the rest of the request.
-        if self.handlerAdded {
-            context.fireChannelRead(data)
-        }
+        context.fireChannelRead(data)
```
Lukasa, line 149: This can fire if adding a handler fails. That's likely to be logically wrong, as this handler will have already written a 500 error, so we mustn't forward it on. The easiest way to tackle that is probably to have `addHandlerOrInternalError` become throwing, and then handle the internal error at the `channelRead` scope.

PR #281, opened 2025-07-17
Add callback in TimedCertificateReloader to listen for reloads

hamzahrmalik:
Add callback in TimedCertificateReloader to listen for reloads

### Motivation:

Adopter might want to know when a certificate was reloaded, e.g. to emit a log or metric

### Modifications:

- Add a new callback in the reloader which is called whenever a reload happens. The closure is passed a struct which tells us which cert and key were read, and whether or not theyre 'new'
- Change the initializer to take a config struct to make this more extensible going forward

### DIFF
--- a/Sources/NIOCertificateReloading/TimedCertificateReloader.swift
+++ b/Sources/NIOCertificateReloading/TimedCertificateReloader.swift
@@ -100,6 +100,21 @@ import Foundation
 /// not being recognized or not matching the configured one; not being able to verify a certificate's signature against the given
 /// private key; etc), then that attempt will be aborted but the service will keep on trying at the configured interval.
 /// The last-valid certificate-key pair (if any) will be returned as the ``sslContextConfigurationOverride``.
+///
+/// Optionally, you may also observe any reloads by specifying the ``TimedCertificateReloader/Configuration/onCertificateLoaded`` parameter.
+/// This will notify you whenever a new certificate is loaded, and you will given an instance of ``TimedCertificateReloader/LoadedCertificateChainAndKeyPairDiff``.
+/// This struct contains the previous certificate and key, as well as the new ones. This is useful for example if you would like to log whenever a new certificate is loaded.
+/// ```swift
+/// let reloaderConfiguration = TimedCertificateReloader.Configuration(
+///     refreshInterval: .seconds(500),
+///     certificateSource: TimedCertificateReloader.CertificateSource(...),
+///     privateKeySource: TimedCertificateReloader.PrivateKeySource(...)
+/// ) { configuration in
+///     configuration.onCertificateLoaded = { diff in
+///         logger.info("Loaded new certificate", metadata: ["certificate": "\(diff.currentX509CertificateChain)"])
+///     }
+/// }
+/// ```
 #if compiler(>=6.0)
 @available(macOS 13, iOS 16, watchOS 9, tvOS 16, macCatalyst 16, visionOS 1, *)
 #else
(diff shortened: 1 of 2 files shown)


Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -278,16 +278,49 @@ public struct TimedCertificateReloader: CertificateReloader {
         }
     }
 
-    private struct CertificateKeyPair {
+    private struct CertificateKeyPair: Hashable {
```
gjcairo, line 281: Unrelated to this change but I wonder if we should rename this struct to `CertificateChainAndKeyPair` to make it clearer that it's not just a single cert.

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -278,16 +278,49 @@ public struct TimedCertificateReloader: CertificateReloader {
...
         var certificates: [NIOSSLCertificateSource]
         var privateKey: NIOSSLPrivateKeySource
     }
 
+    /// Provides information about the newly loaded certificate.
+    public struct LoadedInfo: Sendable {
```
gjcairo, line 287: I think a potentially nicer shape here would be to return both the new and the old pairs instead of having an `isNew` property. Users can then do the comparison themselves, and they also get more information that may be useful.

On that note, I would find something along the lines of `LoadedCertificateChainAndKeyPairDiff` a better name for this type.

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -278,16 +278,49 @@ public struct TimedCertificateReloader: CertificateReloader {
         }
     }
 
-    private struct CertificateKeyPair {
+    private struct CertificateKeyPair: Hashable {
         var certificates: [NIOSSLCertificateSource]
         var privateKey: NIOSSLPrivateKeySource
     }
 
+    /// Provides information about the newly loaded certificate.
```
gjcairo, line 286:
```suggestion
    /// Provides information about the loaded certificate chain and private key.
```

Also, can we add docs for each of the properties?

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -278,16 +278,49 @@ public struct TimedCertificateReloader: CertificateReloader {
...
+    public struct LoadedInfo: Sendable {
+        public var certificates: [Certificate]
+        public var privateKey: Certificate.PrivateKey
+        public var isNew: Bool
+    }
+
+    public struct Configuration: Sendable {
```
gjcairo, line 293: Can we add [a one line] doc for the type?

gjcairo (changes requested):
Thanks Hamzah!

On top of the few comments I've left, don't you think it would also be useful to add a separate callback for whenever a reload fails?

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
hamzahrmalik: maybe in a followup?

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
hamzahrmalik: i agree, updated

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
gjcairo: Sure

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -472,12 +556,25 @@ public struct TimedCertificateReloader: CertificateReloader {
...
+            let oldPair = $0
+            $0 = newPair
+            return oldPair
         }
+        self.onCertificateLoaded?(
+            LoadedCertificateChainAndKeyPairDiff(
+                previousCertificateChain: oldPair?.x509Certificates,
+                previousPrivateKey: oldPair?.x509Key,
+                currentCertificateChain: certificates,
+                currentPrivateKey: key
```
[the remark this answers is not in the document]
hamzahrmalik: the intention of the callback is to allow users to do things such as checking logging, metrics, checking expiry to ensure rotation is working as expected. These things are generally eaier with the x509 types. The intention is not for them to actually _use_ these certs. If they were for using, i would agree they should be niossl types so they can be passed back into things like TLSConfiguration.

Sources/NIOCertificateReloading/TimedCertificateReloader.swift
```diff
@@ -100,6 +100,21 @@ import Foundation
 /// not being recognized or not matching the configured one; not being able to verify a certificate's signature against the given
 /// private key; etc), then that attempt will be aborted but the service will keep on trying at the configured interval.
 /// The last-valid certificate-key pair (if any) will be returned as the ``sslContextConfigurationOverride``.
+///
+/// Optionally, you may also observe any reloads by specifying the ``TimedCertificateReloader/Configuration/onCertificateLoaded`` parameter.
+/// This will notify you whenever a new certificate is loaded, and you will given an instance of ``TimedCertificateReloader/LoadedCertificateChainAndKeyPairDiff``.
```
gjcairo, line 105:
```suggestion
/// This will notify you whenever a new certificate is loaded, and you will be given an instance of ``TimedCertificateReloader/LoadedCertificateChainAndKeyPairDiff``.
```

gjcairo (approved):
LGTM, just one typo in the docs. Also formatter is angry.

PR #286, opened 2025-09-15
Add makeServerConfigurationWithMTLS

hamzahrmalik:
To mirror https://github.com/apple/swift-nio-ssl/pull/545

This allows users to easily create a tls configuration for servers which want to do mtls, with reloading

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,7 @@ jobs:
             linux_5_10_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
             linux_6_0_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
             linux_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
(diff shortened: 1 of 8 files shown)


gjcairo (approved):

Lukasa (commented):
Can you add a test for this code please?

Sources/NIOCertificateReloading/CertificateReloader.swift
```diff
@@ -89,6 +89,43 @@ extension TLSConfiguration {
...
+    /// - Parameters:
+    ///  - certificateReloader: A ``CertificateReloader`` to watch for certificate and key pair updates.
+    ///  - trustRoots: The roots used to validate the client certificate.
+    /// - Returns: A ``NIOSSL/TLSConfiguration`` for use with server-side contexts, that reloads the certificate and key
+    ///   used in its SSL handshake.
+    /// - Throws: This method will throw if an override isn't present. This may happen if a certificate or private key
+    ///   could not be loaded from the given paths.
+    public static func makeServerConfigurationWithMTLS(
+        certificateReloader: some CertificateReloader,
+        trustRoots: NIOSSLTrustRoots
```
gjcairo: Would it make sense to default this to the system defaults?

Lukasa: Yeah, we should not default this to the system trust roots, that will tend to be surprising.

Lukasa (approved):

PR #293, opened 2025-10-31
Handle absolute-form requests when converting from/to HTTPTypes

guoye-zhang:
Handle absolute-form requests when converting from/to HTTPTypes.

### Motivation:

HTTPTypes represent modern HTTP/2 / HTTP/3 semantics which do not support absolute-form requests.

### Modifications:

When converting an absolute-form request to an HTTPRequest, the request target is converted to its path. When converting an HTTPRequest back, the user needs to specify which form to convert to.

### Result:

HTTPRequest no longer contains invalid paths when working with cleartext HTTP/1 proxies.

### DIFF
--- a/Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
+++ b/Sources/NIOHTTPTypesHTTP1/HTTPTypeConversion.swift
@@ -221,8 +221,8 @@ extension HTTPRequestHead {
 extension HTTPRequest {
     private static func requestPath(from oldPath: String) -> String {
         guard let components = URLComponents(string: oldPath),
-              let urlString = components.string,
-              components.rangeOfScheme != nil
+            let urlString = components.string,
+            components.rangeOfScheme != nil
         else {
             return oldPath
         }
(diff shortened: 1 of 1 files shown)


Sources/NIOHTTPTypesHTTP1/HTTPToHTTP1Codec.swift
```diff
@@ -93,14 +93,25 @@ public final class HTTPToHTTP1ServerCodec: ChannelDuplexHandler, RemovableChanne
...
+    /// Initializes a `HTTPToHTTP1ServerCodec`.
+    public init() {
+        self.absoluteForm = false
+    }
+
     /// Initializes a `HTTPToHTTP1ServerCodec`.
-    public init() {}
+    /// - Parameters:
+    ///   - absoluteForm: Whether the request should use the absolute-form (for cleartext HTTP proxies).
+    public init(absoluteForm: Bool) {
```
Lukasa: This change appears to be untested.

guoye-zhang: Added a test

PR #296, opened 2025-11-21
Fix typo in `RPCErrors` case name

H0sungKim:
`fragementHeaderLengthTooShort` -> `fragmentHeaderLengthTooShort`

This pull request fixes a typo in the `RPCErrors` case name, correcting `fragementHeaderLengthTooShort` to `fragmentHeaderLengthTooShort`.
Both the enum case and the call site have been updated accordingly.


Lukasa (changes requested):
Unfortunately, we can't change the name of this enum case because it may be being used by users outside this library, which means this is a semver breaking change.

PR #316, opened 2026-07-02
Throw on invalid psuedoheaders when converting between http/2

glbrntt:
Motivation:

The HTTPType to HTTP2 conversion functions can trip up on invalid psuedo header fields. Prior to this change they'll loop indefinitely, instead they should be rejected per RFC 9113 § 8.3.

Modifications:

- Throw on unrecognised psuedo header fields
- Add tests

Result:

- More spec compliant

### DIFF
--- a/Tests/NIOHTTPTypesHTTP2Tests/NIOHTTPTypesHTTP2Tests.swift
+++ b/Tests/NIOHTTPTypesHTTP2Tests/NIOHTTPTypesHTTP2Tests.swift
@@ -190,6 +190,7 @@ final class NIOHTTPTypesHTTP2Tests: XCTestCase {
         try self.channel.writeInbound(HTTP2Frame.FramePayload(headers: malformedHeaders))
         XCTAssertNil(try self.channel.readInbound(as: HTTPRequestPart.self))
         XCTAssertEqual(errorRecorder.caughtErrors.count, 1)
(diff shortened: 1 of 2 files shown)


Tests/NIOHTTPTypesHTTP2Tests/NIOHTTPTypesHTTP2Tests.swift
```diff
@@ -162,4 +174,65 @@ final class NIOHTTPTypesHTTP2Tests: XCTestCase {
...
+        // `HTTP2TypeConversionError.unknownPseudoField` should be thrown if an invalid psuedo header
+        // is included in the response.
+        let headers: HPACKHeaders = [
+            ":method": "GET",
+            ":scheme": "https",
+            ":path": "/",
+            ":status": "200",  // invalid for requests
+        ]
+
+        XCTAssertThrowsError(try HTTPRequest(headers))
```
rafaelcepeda: For this test and the other ones, shouldn't we try to assert that the specific `HTTP2TypeConversionError.unknownPseudoField` was thrown? At the moment, we are only asserting that _an_ error was thrown.

rafaelcepeda (commented):
Left a comment about how we are testing this.

rafaelcepeda (approved):
LGTM

PR #317, opened 2026-07-03
Make the HTTP decompressors safe against synchronous channelRead re-entry

### LINKED ISSUE #314: Decompressor.channelRead holds an exclusive access to self across fireChannelRead/fireErrorCaught → exclusivity crash (SIGABRT) on synchronous re-entry
### Summary

`NIOHTTPResponseDecompressor` (and its twin `NIOHTTPRequestDecompressor`) keep their state in a stored `var decompressor: Decompressor<…>` (a struct) and forward to its **`mutating func channelRead`**. That mutating call holds an **exclusive (modify) access to `self.decompressor` for its entire duration**, and it calls `context.fireChannelRead(_:)` / `context.fireErrorCaught(_:)` *while that access is still open*.

If anything downstream **synchronously re-enters the same handler's `channelRead`** before the outer call returns, the re-entrant call opens a *second* exclusive access to `self.decompressor` and the Swift runtime aborts with:

```
Fatal access conflict detected.
Simultaneous accesses to 0x…, but modification requires exclusive access.
```

We hit this reliably in production via **AsyncHTTPClient over HTTP/2**: a decompression error closes the HTTP/2 stream synchronously, which re-delivers a buffered read into the same decompressor while it is mid-`channelRead`.

### Where the unsafe pattern is (swift-nio-extras 1.34.1)

`Sources/NIOHTTPCompression/HTTPDecompressor.swift`:

```swift
public final class NIOHTTPResponseDecompressor: ChannelDuplexHandler, ... {
    private var decompressor: Decompressor<HTTPResponseHead>          // stored var (struct)

    public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
        self.decompressor.channelRead(context: context, part: Self.unwrapInboundIn(data))   // mutating → exclusive access to self.decompressor
    }
}

private struct Decompressor<InboundHead: HTTPHead> {
    mutating func channelRead(context: ChannelHandlerContext, part: Inbound) {
        switch part {
        case .head:
            ...
            context.fireChannelRead(NIOAny(part))          // fired while the mutating access is OPEN
        case .body(var content):
            ...
            context.fireChannelRead(NIOAny(Inbound.body(buffer)))   // ditto
            ...
            } catch {
                context.fireErrorCaught(error)             // ditto  ← our crash path
            }
        case .end:
            ...
            context.fireChannelRead(NIOAny(part))          // ditto
        }
    }
}
```

Every `context.fire…` on lines ~129/131/135/151/158/161/172/175 happens inside the `mutating func`, i.e. while the exclusive access to `self.decompressor` is held.

### How AsyncHTTPClient triggers it

AHC 1.33.1 installs the decompressor into the connection pipeline:

- HTTP/2: `Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift:279` → `NIOHTTPResponseDecompressor(limit: limit)`
- HTTP/1: `Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1Connection.swift:146`

Sequence (from the symbolized core, bottom→top):

```
NIOHTTPCompression … Decompressor.channelRead            ← OUTER call, exclusive access to self.decompressor opens
  → context.fireErrorCaught(error)                        (decompress() failed, e.g. limit exceeded)
AsyncHTTPClient.HTTP2ClientRequestHandler.runFailedFinalAction
  → HTTP2 stream close (HTTP2StreamChannel.receiveStreamClosed)
  → pending reads re-delivered: HTTP2FramePayloadToHTTP1ClientCodec.channelRead → ChannelHandlerContext.invokeChannelRead
NIOHTTPCompression … Decompressor.channelRead            ← INNER call, SECOND exclusive access to self.decompressor
swift_beginAccess
swift::runtime::AccessSet::insert(..., ExclusivityFlags)
swift::fatalError → abort()   (SIGABRT)
```

(The top frame symbolized as `NIOHTTPRequestDecompressor`; AHC installs `NIOHTTPResponseDecompressor`. They share the same `Decompressor` struct, so the defect is identical; the symbol is the adjacent generic twin.)

### Reproduction

The essential condition is: **`Decompressor.channelRead` is re-entered synchronously while its outer invocation is still on the stack.** Minimal `EmbeddedChannel` reproducer below — a downstream handler synchronously re-injects an inbound read when it first sees one, which re-enters the decompressor and trips exclusivity. (Any build with dynamic exclusivity enforcement on, i.e. the default; running it aborts the process.)

<details>
<summary>Reproducer (drop into <code>Tests/NIOHTTPCompressionTests/</code>)</summary>

```swift
import NIOCore
import NIOEmbedded
import NIOHTTP1
import XCTest
@testable import NIOHTTPCompression

final class DecompressorReentrancyTests: XCTestCase {
    /// Synchronously re-injects an inbound read the first time it sees one — mimicking
    /// what AsyncHTTPClient's HTTP/2 stream-close path does (deliverPendingReads) while
    /// the decompressor's `mutating func channelRead` is still on the stack.
    final class ReentrantOnRead: ChannelInboundHandler {
        typealias InboundIn = HTTPClientResponsePart
        private var fired = false
        private let replay: HTTPClientResponsePart
        init(replay: HTTPClientResponsePart) { self.replay = replay }
        func channelRead(context: ChannelHandlerContext, data: NIOAny) {
            if !fired { fired = true; context.channel.pipeline.fireChannelRead(NIOAny(replay)) }
            context.fireChannelRead(data)
        }
    }

    func testSynchronousReentryDuringDecompressorChannelReadTrapsExclusivity() throws {
        var head = HTTPResponseHead(version: .http1_1, status: .ok)
        head.headers.add(name: "Content-Encoding", value: "gzip")
        let replay = HTTPClientResponsePart.head(head)

        try channel.pipeline.syncOperations.addHandler(NIOHTTPResponseDecompressor(limit: .ratio(25)))
        try channel.pipeline.syncOperations.addHandler(ReentrantOnRead(replay: replay))

        // First .head enters NIOHTTPResponseDecompressor.channelRead -> the struct's
        // `mutating func channelRead` (exclusive access to self.decompressor opens) ->
        // fireChannelRead(.head) downstream -> ReentrantOnRead re-injects another .head ->
        // decompressor.channelRead re-enters -> SECOND exclusive access -> abort (SIGABRT).
        try channel.writeInbound(replay)
        _ = try? channel.finish()
    }
}
```
</details>

(Not yet compiled inside the package — written against 1.34.1; adjust symbols if needed. The mechanism, not the exact bytes, is the point.)

In production it reproduces with:
- `HTTPClient.Configuration.decompression = .enabled(limit: .ratio(25))`
- HTTP/2 origins returning compressed bodies, under load (the trigger is the decompression error path closing the stream synchronously; large/very-compressible bodies that exceed the ratio limit make it frequent).

### Environment

- swift-nio-extras **1.34.1**
- async-http-client **1.33.1**, swift-nio **2.99.0**, swift-nio-http2 **1.44.0**, swift-nio-ssl **2.37.1**
- Swift 6.3.2, Ubuntu 24.04 LTS, aarch64 (Raspberry Pi 5)

### Suggested fix (for discussion)

The handler should not call `context.fireChannelRead`/`fireErrorCaught` while holding an exclusive access to `self.decompressor`. Options:

1. **Drain, then fire.** Have `Decompressor.channelRead` produce the outputs (decompressed buffers + any error/forwarded parts) into a local array, return it from the mutating call, and let the *handler* fire them downstream after the mutating access has ended. No `context.fire…` happens inside the exclusive-access window, so synchronous re-entry is safe.
2. **Re-entrancy guard / buffering** in the handler: if `channelRead` is re-entered while already active, enqueue the part and process it after the current call unwinds.
3. (AHC side, complementary) On stream failure/close, avoid synchronously re-delivering reads into a handler that is currently executing `channelRead`; defer with `context.eventLoop.execute { … }`.

I think (1) is the most localized and robust, and I'm happy to put up a PR if you agree on the approach.

MichaelDRZ:
### Motivation:

`NIOHTTPRequestDecompressor` / `NIOHTTPResponseDecompressor` keep their state in a
stored `var decompressor` (a `struct`) and forwarded downstream (`fireChannelRead` /
`fireErrorCaught`) from inside that struct's `mutating` `channelRead`, i.e. while an
exclusive (modify) access to the stored property was still open. If firing downstream
synchronously re-enters the handler's `channelRead`, a second exclusive access is
opened and the program traps with `Fatal access conflict detected` (SIGABRT).

This reproduces in production with AsyncHTTPClient over HTTP/2 and
`decompression = .enabled`: a compressed response is mid-decompress when a
decompression error closes the stream, and the buffered read re-delivered on the
stream-close path re-enters `channelRead` on the same decompressor.

### Modifications:

- The body decompression is now expressed as a `NIOSingleStepByteToMessageDecoder`,
  driven by a `NIOSingleStepByteToMessageProcessor`. The handler pulls one decompressed
  buffer at a time and forwards it only after `decode` has returned, so the decoder
  relinquishes its exclusive access before anything is fired downstream.
- The shared driver is now a `class`, so the head/end forwarding paths are re-entrancy
  safe for the same reason.
- The zlib stream is torn down in the decoder's `deinit`, which additionally fixes a
  leak when a message ends without a final `.end`.
- No public API changes.

### Result:

Synchronous re-entry of `channelRead` no longer traps. Adds regression tests covering
the forward and error re-entry paths on both decompressors; all existing decompression
tests (limits, trailing data, truncation, multi-message reuse) continue to pass.

Fixes #314

### DIFF
--- a/Sources/NIOHTTPCompression/HTTPDecompressor.swift
+++ b/Sources/NIOHTTPCompression/HTTPDecompressor.swift
@@ -93,83 +93,117 @@ private protocol HTTPHead: Equatable {
 extension HTTPRequestHead: HTTPHead {}
 extension HTTPResponseHead: HTTPHead {}
 
-private struct Decompressor<InboundHead: HTTPHead> {
+/// This class encapsulates the state of a single http message decompression
+private final class HTTPDecompressionDecoder: NIOSingleStepByteToMessageDecoder {
+    typealias InboundOut = ByteBuffer
+
+    private let allocator: ByteBufferAllocator
+    private var decompressor: NIOHTTPDecompression.Decompressor
+    /// The number of already consumed compressed bytes
+    private var compressedLength: Int
+    private var complete: Bool
+
+    init(limit: NIOHTTPDecompression.DecompressionLimit, allocator: ByteBufferAllocator) throws {
+        self.allocator = allocator
+        self.decompressor = NIOHTTPDecompression.Decompressor(limit: limit)
+        self.compressedLength = 0
+        self.complete = false
+        try self.decompressor.initializeDecoder()
+    }
+
+    deinit {
+        self.decompressor.deinitializeDecoder()
+    }
+
+    func decode(buffer: inout ByteBuffer) throws -> ByteBuffer? {
+        guard !self.complete, buffer.readableBytes > 0 else {
+            return nil
+        }
+
+        let compressedLength = self.compressedLength + buffer.readableBytes
+        var output = self.allocator.buffer(capacity: 16384)
+        let readableBytesBefore = buffer.readableBytes
+        let result = try self.decompressor.decompress(
+            part: &buffer,
+            buffer: &output,
+            compressedLength: compressedLength
+        )
+        self.compressedLength += readableBytesBefore - buffer.readableBytes
+        if result.complete {
+            self.complete = true
+        }
+        return output
+    }
+
+    func decodeLast(buffer: inout ByteBuffer, seenEOF: Bool) throws -> ByteBuffer? {
+        if let output = try self.decode(buffer: &buffer) {
+            return output
+        }
+        if !self.complete {
+            throw NIOHTTPDecompression.ExtraDecompressionError.truncatedData
+        }
+        return nil
+    }
+}
+
+private final class Decompressor<InboundHead: HTTPHead> {
     typealias Inbound = HTTPPart<InboundHead, ByteBuffer>
 
-    /// this struct encapsulates the state of a single http response decompression
-    private struct Compression {
-        /// the used algorithm
-        var algorithm: NIOHTTPDecompression.CompressionAlgorithm
-
-        /// the number of already consumed compressed bytes
-        var compressedLength: Int
-    }
-
-    private var compression: Compression? = nil
-    private var decompressor: NIOHTTPDecompression.Decompressor
-    private var decompressionComplete: Bool
+    private let limit: NIOHTTPDecompression.DecompressionLimit
+    private var processor: NIOSingleStepByteToMessageProcessor<HTTPDecompressionDecoder>?
 
     init(limit: NIOHTTPDecompression.DecompressionLimit) {
-        self.decompressor = NIOHTTPDecompression.Decompressor(limit: limit)
-        self.decompressionComplete = false
+        self.limit = limit
+        self.processor = nil
     }
 
-    mutating func channelRead(context: ChannelHandlerContext, part: Inbound) {
+    func channelRead(context: ChannelHandlerContext, part: Inbound) {
         switch part {
         case .head(let head):
             let contentType = head.headers[canonicalForm: "Content-Encoding"].first?.lowercased()
             let algorithm = NIOHTTPDecompression.CompressionAlgorithm(header: contentType)
 
             do {
-                if let algorithm = algorithm {
-                    self.compression = Compression(algorithm: algorithm, compressedLength: 0)
-                    try self.decompressor.initializeDecoder()
+                if algorithm != nil {
+                    let decoder = try HTTPDecompressionDecoder(
+                        limit: self.limit,
+                        allocator: context.channel.allocator
+                    )
+                    self.processor = NIOSingleStepByteToMessageProcessor(decoder)
                 }
 
                 context.fireChannelRead(NIOAny(part))
             } catch {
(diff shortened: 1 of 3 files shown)


glbrntt (changes requested):
Thanks – this looks great! You just need to appease the formatter.

glbrntt (approved):
Great patch; thanks for fixing this!

PR #319, opened 2026-07-07
Fix HTTPResumableUploadHandler dropping responses after the first request on a connection

### LINKED ISSUE #318: HTTPResumableUploadHandler drops all responses after the first request on a connection
### Expected behavior

A server pipeline containing `HTTPResumableUploadHandler` can serve multiple requests on the same connection (HTTP/1.1 keep-alive). Each request gets its response.

### Actual behavior

Only the first request gets a response. For the second and every following request on the same channel, response writes are silently dropped, their promises never complete, and `read()` is no longer forwarded to the parent channel. The client hangs until it times out. No error is surfaced anywhere.

Cause: the handler routes all outbound ops through its stored `context`, which is set only in `handlerAdded`. When the second request head arrives, `channelRead` calls `resetUpload`, which ends the previous upload. That synchronously runs `detachUploadHandler` → `handler.detach()`, which sets `self.context = nil` — and nothing sets it again. From then on every `self.context?.write(...)` is a no-op.

This doesn't show up over HTTP/2 (fresh handler per stream) and the existing tests all use a fresh `EmbeddedChannel` per request, which is why it went unnoticed.

### Steps to reproduce

1. Add `HTTPResumableUploadHandler` to a server pipeline.
2. Send two requests over one connection (plain GETs are enough, no resumable upload needed).
3. First request is answered, second one hangs.

### If possible, minimal yet complete reproducer code (or URL to code)

```swift
final class OKResponder: ChannelDuplexHandler {
    typealias InboundIn = HTTPRequestPart
    typealias OutboundIn = Never
    typealias OutboundOut = HTTPResponsePart

    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
        if case .end = self.unwrapInboundIn(data) {
            context.write(self.wrapOutboundOut(.head(HTTPResponse(status: .ok))), promise: nil)
            context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
        }
    }
}

let context = HTTPResumableUploadContext(origin: "https://example.com")
try channel.pipeline.syncOperations.addHandler(
    HTTPResumableUploadHandler(context: context, handlers: [OKResponder()])
)

let request = HTTPRequest(method: .get, scheme: "https", authority: "example.com", path: "/")

try channel.writeInbound(HTTPRequestPart.head(request))
try channel.writeInbound(HTTPRequestPart.end(nil))
print(try channel.readOutbound(as: HTTPResponsePart.self)) // .head(200 OK)
print(try channel.readOutbound(as: HTTPResponsePart.self)) // .end

try channel.writeInbound(HTTPRequestPart.head(request))
try channel.writeInbound(HTTPRequestPart.end(nil))
print(try channel.readOutbound(as: HTTPResponsePart.self)) // nil
print(try channel.readOutbound(as: HTTPResponsePart.self)) // nil
```

Fix is a one-liner: restore `self.context = context` in `resetUpload`. PR incoming.

main, 88a51340f59cf181ebde888bd1b749296b3ec029

```
swift-driver version: 1.148.6 Apple Swift version 6.3.3 (swiftlang-6.3.3.1.3 clang-2100.1.1.101)
Target: arm64-apple-macosx26.0
Darwin Mac 25.5.0 Darwin Kernel Version 25.5.0: Tue Jun  9 22:28:24 PDT 2026; root:xnu-12377.121.10~1/RELEASE_ARM64_T6020 arm64
```

MichaelDRZ:
Fix `HTTPResumableUploadHandler` silently dropping all responses after the first request on a connection.

### Motivation:

`HTTPResumableUploadHandler` forwards all outbound operations through its stored `ChannelHandlerContext`, which is set only in `handlerAdded` and cleared by `detach()` whenever an upload releases the handler. When a second request arrives on the same channel — the normal case for HTTP/1.1 with keep-alive — `channelRead` calls `resetUpload`, which ends the previous upload. That synchronously detaches the handler and clears its context, and nothing restores it before the new upload is attached. Every response write for the second and subsequent requests then hits `self.context?.write(...)` as a no-op: the response is silently dropped, write promises never complete, and `read()` forwarding to the parent channel stops, so the client hangs until it times out.

The existing tests never caught this because each test sends one request per `EmbeddedChannel`; resumption is only exercised across separate channels, and per-stream channels (HTTP/2) recreate the handler per request, which also masks the bug.

### Modifications:

- Restore `self.context` in `resetUpload(context:)` after ending the previous upload.
- Add `testNonUploadKeepAlive` and `testResumableUploadKeepAlive`, which send a second request on the same `EmbeddedChannel` for pass-through traffic and after a completed upload creation respectively. Both fail without the fix.

### Result:

Connections handled by `HTTPResumableUploadHandler` can serve more than one request: responses to the second and subsequent requests on a keep-alive connection are delivered instead of being silently dropped.

Fixes #318

### DIFF
--- a/Sources/NIOResumableUpload/HTTPResumableUploadHandler.swift
+++ b/Sources/NIOResumableUpload/HTTPResumableUploadHandler.swift
@@ -70,6 +70,7 @@ public final class HTTPResumableUploadHandler: ChannelDuplexHandler {
         if let existingUpload = self.upload {
             existingUpload.end(handler: self, error: nil)
         }
+        self.context = context
         let upload = self.createUpload()
         upload.scheduleOnEventLoop(self.eventLoop)
         upload.attachUploadHandler(self.sendableView, channel: context.channel)
--- a/Tests/NIOResumableUploadTests/NIOResumableUploadTests.swift
+++ b/Tests/NIOResumableUploadTests/NIOResumableUploadTests.swift
@@ -84,6 +84,125 @@ final class NIOResumableUploadTests: XCTestCase {
         XCTAssertTrue(try channel.finish().isClean)
     }
 
+    func testNonUploadKeepAlive() throws {
+        let channel = EmbeddedChannel()
+        let recorder = InboundRecorder<HTTPRequestPart, HTTPResponsePart>()
+
+        let context = HTTPResumableUploadContext(origin: "https://example.com")
+        try channel.pipeline.syncOperations.addHandler(
+            HTTPResumableUploadHandler(context: context, handlers: [recorder])
+        )
+
+        let request = HTTPRequest(method: .get, scheme: "https", authority: "example.com", path: "/")
+        try channel.writeInbound(HTTPRequestPart.head(request))
+        try channel.writeInbound(HTTPRequestPart.end(nil))
+
+        XCTAssertEqual(recorder.receivedFrames.count, 2)
+
+        recorder.write(HTTPResponsePart.head(HTTPResponse(status: .ok)))
+        recorder.write(HTTPResponsePart.end(nil))
+
+        let responsePart = try channel.readOutbound(as: HTTPResponsePart.self)
+        guard case .head(let response) = responsePart else {
+            XCTFail("Part is not response headers")
+            return
+        }
+        XCTAssertEqual(response.status, .ok)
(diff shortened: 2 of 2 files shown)


Lukasa (approved):
This is a very nice change, thank you very much! :sparkles:

PR #321, opened 2026-07-31
Keep zlib streams at stable addresses

### LINKED ISSUE #308: Wrap z_stream in an UnsafeMutablePointer
> zlib depends on the address of the z_stream struct being unchanged between calls, so storing it directly is error prone. And although it might currently be working as expected, it relies on undefined behavior, see: https://github.com/swiftlang/swift/blob/af3e7e765549c0397288e60983c96d81639287ed/stdlib/public/core/UnsafePointer.swift#L197-L205
> 
> had a similar issue here: https://github.com/adam-fowler/compress-nio/pull/12 

 _Originally posted by @Cyberbeni in [#306](https://github.com/apple/swift-nio-extras/pull/306/changes#r2954613173)_

thliu21:
### Motivation:

zlib stores the address of `z_stream` in its internal state and expects that
address to remain stable across calls. Storing `z_stream` inline in movable
Swift value types does not provide that guarantee and can cause
`Z_STREAM_ERROR` during cleanup.

Fixes #308.

### Modifications:

- Make the internal compressor and decompressor reference owners of separately
  allocated `z_stream` storage.
- Pass the stable pointer to every zlib initialization, processing, and cleanup
  call.
- Keep lifecycle state with the pointer owner and make conditional cleanup
  idempotent.
- Add regression tests covering failed initialization recovery, multiple
  compression and decompression calls, repeated lifecycles, and deallocation
  cleanup.

### Result:

zlib observes the same `z_stream` address for each complete lifecycle, while
stream cleanup remains single-owner and safe against repeated shutdown.

### DIFF
--- a/Tests/NIOHTTPCompressionTests/ZlibStreamLifecycleTest.swift
+++ b/Tests/NIOHTTPCompressionTests/ZlibStreamLifecycleTest.swift
@@ -85,21 +85,4 @@ final class ZlibStreamLifecycleTest: XCTestCase {
         }
     }
 
-    func testActiveStreamsAreCleanedUpOnDeinit() throws {
-        weak var weakCompressor: NIOCompression.Compressor?
-        weak var weakDecompressor: NIOHTTPDecompression.Decompressor?
-
-        do {
-            let compressor = NIOCompression.Compressor()
-            compressor.initialize(encoding: .gzip)
-            weakCompressor = compressor
-
-            let decompressor = NIOHTTPDecompression.Decompressor(limit: .none)
-            try decompressor.initializeDecoder()
-            weakDecompressor = decompressor
-        }
-
-        XCTAssertNil(weakCompressor)
-        XCTAssertNil(weakDecompressor)
-    }
 }
--- a/Tests/NIOHTTPCompressionTests/ZlibStreamLifecycleTest.swift
+++ b/Tests/NIOHTTPCompressionTests/ZlibStreamLifecycleTest.swift
@@ -45,6 +45,7 @@ final class ZlibStreamLifecycleTest: XCTestCase {
             (NIOCompression.Algorithm.deflate, ["second compression ", "lifecycle"]),
         ] {
             compressor.initialize(encoding: encoding)
+            XCTAssertTrue(compressor.zlibStateIsAllocated)
             var firstInput = allocator.buffer(capacity: chunks[0].utf8.count)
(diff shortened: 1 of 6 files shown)


Tests/NIOHTTPCompressionTests/ZlibStreamLifecycleTest.swift
```diff
@@ -0,0 +1,101 @@
...
+            weakCompressor = compressor
+
+            let decompressor = NIOHTTPDecompression.Decompressor(limit: .none)
+            try decompressor.initializeDecoder()
+            weakDecompressor = decompressor
+        }
+
+        XCTAssertNil(weakCompressor)
+        XCTAssertNil(weakDecompressor)
+    }
```
glbrntt: I don't think tests proves anything useful: the compressor/decompressor are classes so it just shows that `weak` references work. It doesn't prove that the `z_stream`s are actually cleared up.

glbrntt (changes requested):

thliu21:
Thanks — I update the lifecycle coverage. It now asserts that the underlying z_stream.state is allocated after initialization and cleared after deflateEnd/inflateEnd, rather than only checking that the Swift owner is deallocated. The targeted ZlibStreamLifecycleTest passes locally.

Tests/NIOHTTPCompressionTests/ZlibStreamLifecycleTest.swift
```diff
@@ -0,0 +1,105 @@
...
+            compressor.initialize(encoding: .gzip)
+            weakCompressor = compressor
+
+            let decompressor = NIOHTTPDecompression.Decompressor(limit: .none)
+            try decompressor.initializeDecoder()
+            weakDecompressor = decompressor
+        }
+
+        XCTAssertNil(weakCompressor)
+        XCTAssertNil(weakDecompressor)
```
glbrntt: This test still doesn't verify what it claims to so should be removed

glbrntt (changes requested):

thliu21:
Thanks — I removed `testActiveStreamsAreCleanedUpOnDeinit()` as requested.

Locally, `ZlibStreamLifecycleTest` passes (2 tests), the full `NIOHTTPCompressionTests` suite passes (81 tests), and the explicit-Sendable build and strict format lint both pass.

glbrntt (approved):
Great, thank you!

PR #322, opened 2026-08-24
Make TimedCertificateReloader reload asynchronously

### LINKED ISSUE #297: TimedCertificateReloader is using blocking I/O instead of NIOFileSystem
https://github.com/apple/swift-nio-extras/blob/b87fdbf492c8fd5ac860e642c714d2d24156990a/Sources/NIOCertificateReloading/TimedCertificateReloader.swift#L564

The `public func reload() throws` is clearly not right, it needs to be `public func reload() async throws` because we cannot guarantee that the file I/O is synchronous. In fact in the places where this is useful, it's quite likely that the orchestrator is using a virtualised networked filesystem which is very much not synchronous.

The whole implementation needs to be vetted for `Foundation` usage which does blocking I/O. If (for the time being whilst NIOFileSystem is still `_NIOFileSystem) we need a temporary solution, then the blocking calls need to absolutely be offloaded to a `NIOThreadPool` to be guaranteed not to run on the EventLoops or the Swift Concurrency threads

thliu21:
Make timed certificate reloads asynchronous without blocking Swift concurrency executors.

### Motivation:

`TimedCertificateReloader` currently loads certificate and private-key sources synchronously. File-backed sources may live on virtualized or networked filesystems, so performing this work directly can block an EventLoop or Swift concurrency executor.

Fixes #297.

### Modifications:

- Add async throwing overloads of `reload()` and `makeReloaderValidatingSources`, while deprecating the synchronous overloads.
- Run certificate and private-key source loading on `NIOThreadPool.singleton`.
- Keep parsing, state updates, and user callbacks outside the blocking thread pool.
- Make periodic reloads use the async path and propagate task cancellation.
- Add a regression test covering the source-loading and callback execution contexts.
- Add the explicit `NIOPosix` target dependency required for `NIOThreadPool`.

### Result:

Callers can migrate to an async API that keeps source loading off EventLoops and Swift concurrency threads. Existing synchronous call sites remain source-compatible with deprecation warnings.

### Testing:

- `swift test --filter TimedCertificateReloaderTests` — 25 tests passed on the final commit.
- Strict `NIOCertificateReloading` target build with warnings-as-errors, explicit sendability, and explicit dependency checking.
- Strict Swift format lint and `git diff --check`.

Local validation was performed on arm64 macOS. Linux, the repository Static SDK lane, Windows, and WASI were not run locally. An earlier near-final revision passed the full package suite, but that suite was not rerun against the exact final commit.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -271,6 +271,7 @@ var targets: [PackageDescription.Target] = [
         name: "NIOCertificateReloading",
         dependencies: [
             .product(name: "NIOCore", package: "swift-nio"),
+            .product(name: "NIOPosix", package: "swift-nio"),
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "X509", package: "swift-certificates"),
             .product(name: "SwiftASN1", package: "swift-asn1"),
--- a/Sources/NIOCertificateReloading/TimedCertificateReloader.swift
+++ b/Sources/NIOCertificateReloading/TimedCertificateReloader.swift
@@ -36,7 +36,7 @@ import Foundation
 /// key via ``init(refreshInterval:certificateSource:privateKeySource:logger:)``.
 /// Simply creating a timed reloader won't validate that the sources provide valid certificate and private key pairs. If you want this to be
 /// validated at creation time, you may instead use
-/// ``makeReloaderValidatingSources(refreshInterval:certificateSource:privateKeySource:logger:)``.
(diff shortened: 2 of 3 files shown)


thliu21:
I disambiguated the DocC references so they explicitly link to the new async overloads. The exact documentation command now passes locally, along with all 25 targeted TimedCertificateReloaderTests, strict format lint, and git diff --check. The remaining semantic-version check requires a semver label.

PR #323, opened 2026-09-03
Specify the defining module for HTTP2ErrorCode

glbrntt:
Motivation:

swift-http-types 1.7.0 adds an HTTP2ErrorCode type which is ambiguous with the type of the same name from swift-nio-http2. This causes compilation failures when users depend on the swift-nio-extras.

Modifications:

- Disambiguate the type

Result:

Build compiles


simonjbeaumont (approved):

gjcairo (approved):

fabianfett (approved):

glbrntt:
API breakage check is failing because it can't build `main` for the reason that this PR fixes, merging over it.

